Issue #592, opened 2021-03-16
Labels of colorbars should be next to colorbars, not on top


AntoineSIMTEK:
![grafik](https://user-images.githubusercontent.com/40853209/111287639-b60a6f00-8643-11eb-99ba-272af261569a.png)

That is my opinion. What do you think @pastewka

pastewka:
Ageed, the labels should be next to the colorbar.

I never realized this, but the units are bogus. The units here should be um. We also should *under all circumstance* avoid the e Notation. Please change this to 10^-4. This has a high priority!

AntoineSIMTEK:
I think it depends wether I give the sizes as 0.0005 m or 0.5 mm


![grafik](https://user-images.githubusercontent.com/40853209/111289641-a3913500-8645-11eb-9df6-683d15449cfb.png)

AntoineSIMTEK:
I find this actually fair. That way the users is able to choose his prefered unit

mcrot:
I'm currently implementing this. Can we write "Contact pressure p(E*)" shorter without loosing clarity?
For example: "Pressure p(E*)" with "Contact"?

Otherwise I cannot currently only make it fit by shifting the tick labels of the colorbar much more to the right:
![grafik](https://user-images.githubusercontent.com/9113541/115053134-1577cb00-9edf-11eb-94c3-13f02b0e5bb8.png)

I think the color title should also be shorter in order to make the plot more compact.

mcrot:
Okay, I'm using "Pressure p(E*)".

AntoineSIMTEK:
What we usually do is to rotate the title of the colorbar, like the y label.

mcrot:
Okay, thanks. I'm not sure whether bokeh allows this, let me see.

Something else: What are the units of "pressure" here? Isn't it also normalized? 
The current version shows no unit at all for the pressure plot.

Shouldn't we say "Normalized pressure p/E*" here, like on the x axis on the left side?  Or is the unit missing?

pastewka:
The unit is "E*"

mcrot:
> What we usually do is to rotate the title of the colorbar, like the y label.

Although there are several properties to change the title of a colorbar, there is no easy way to write it vertically.
It is maybe possible by adding a third "column" to the plot which only contains the text, but I guess that
would cause some problems for calculating the correct width of the plot. I would like to leave it above the colorbar at the moment, unless it had a high priority to display it vertically.

[...]

PR #645, opened 2021-04-28
GUI Cosmetics

### LINKED ISSUE #587: Rename "Topography" to "Measurement"
What we call a "topography" should be changed into a "measurements" in the user interface. This more clearly describes what it is.

### LINKED ISSUE #634: Adjust layout of contact calculations to use pills on the left

### LINKED ISSUE #649: Wrong orientation of contact geometry plot
I've generated a simple area measurement with a gradient in `y` direction and constant in `x` direction.
See [topo_2D.txt](https://github.com/ContactEngineering/TopoBank/files/6405192/topo_2D.txt).

In the web app it looks like this:

grafik

The contact geometry has this orientation, which is wrong:
grafik

pastewka:
* Added a topobank-info-box style to be used for information context
* Renamed "Topography" to "Measurement", closes #587 
* Updated layout of contact mechanics calculations, now uses pills on the left, closes #634 
* Changed font size for Bokeh plot labels
* Fixed orientation of contact mechanics plots, closes #649

### DIFF
--- a/topobank/analysis/views.py
+++ b/topobank/analysis/views.py
@@ -29,6 +29,7 @@ from bokeh.plotting import figure
 from bokeh.embed import components, json_item
 from bokeh.models.widgets import CheckboxGroup, Tabs, Panel, Toggle, Div, Slider
 from bokeh.models import LinearColorMapper, ColorBar
+from bokeh.models.formatters import FuncTickFormatter
 
 import xarray as xr
 
--- a/topobank/analysis/views.py
+++ b/topobank/analysis/views.py
@@ -683,14 +683,14 @@ class PlotCardView(SimpleCardView):
                 visible=False,
                 active=list(range(len(subject_checkbox_groups[topography_ct]))))
 
-        subject_btn_group_toggle_button_label = "Topographies"
+        subject_btn_group_toggle_button_label = "Measurements"
         if has_at_least_one_surface_subject:
             subject_btn_group_toggle_button_label = "Average / "+subject_btn_group_toggle_button_label
         subject_btn_group_toggle_button = Toggle(label=subject_btn_group_toggle_button_label)
-        series_btn_group_toggle_button = Toggle(label="Data Series")
-        options_group_toggle_button = Toggle(label="Plot Options")
+        series_btn_group_toggle_button = Toggle(label="Data series")
+        options_group_toggle_button = Toggle(label="Plot options")
 
-        topography_alpha_slider = Slider(start=0, end=1, title="Opacity of topography lines",
+        topography_alpha_slider = Slider(start=0, end=1, title="Opacity of measurement lines",
                                          value=DEFAULT_ALPHA_FOR_TOPOGRAPHIES if has_at_least_one_surface_subject else 1.0,
                                          step=0.1, visible=False)
         options_group = column([topography_alpha_slider])
@@ -915,7 +915,7 @@ class ContactMechanicsCardView(SimpleCardView):
                 visible=False,
                 active=list(range(len(topography_names))))  # all active
 
-            topography_btn_group_toggle_button = Toggle(label="Topographies")
+            topography_btn_group_toggle_button = Toggle(label="Measurements")
 
             # extend mapping of Python to JS objects
             js_args['topography_btn_group'] = topography_button_group
@@ -968,7 +968,7 @@ class ContactMechanicsCardView(SimpleCardView):
                  message="""
                  Translucent data points did not converge within iteration limit and may carry large errors.
                  <i>A</i> is the true contact area and <i>A0</i> the apparent contact area,
-                 i.e. the size of the provided topography.""")
+                 i.e. the size of the provided measurement.""")
         ]
 
         context['limits_calc_kwargs'] = settings.CONTACT_MECHANICS_KWARGS_LIMITS
@@ -1099,7 +1099,7 @@ def submit_analyses_view(request):
     return JsonResponse({}, status=status)
 
 
-def _contact_mechanics_geometry_figure(values, frame_width, frame_height, topo_unit, topo_size, title=None,
+def _contact_mechanics_geometry_figure(values, frame_width, frame_height, topo_unit, topo_size, colorbar_title=None,
                                        value_unit=None):
     """
 
@@ -1108,7 +1108,7 @@ def _contact_mechanics_geometry_figure(values, frame_width, frame_height, topo_u
     :param frame_height:
     :param topo_unit:
     :param topo_size:
-    :param title:
+    :param colorbar_title:
     :param value_unit:
     :return:
     """
@@ -1150,7 +1150,7 @@ def _contact_mechanics_geometry_figure(values, frame_width, frame_height, topo_u
                             label_standoff=COLORBAR_LABEL_STANDOFF,
                             width=COLORBAR_WIDTH,
                             location=(0, 0),
-                            title=f"{title} ({value_unit})")
+                            title=f"{colorbar_title} ({value_unit})")
         colorbar.formatter = FuncTickFormatter(code="return format_exponential(tick);")
         p.add_layout(colorbar, "right")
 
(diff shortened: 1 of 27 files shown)


pastewka:
Hi @mcrot - there is an issue with this that I just opened #646 - can you give advise how to resolve it?

pastewka:
Hi @mcrot - I think I figured it out... but it needs your review

topobank/templates/manager/surface_detail.html
```diff
@@ -99,13 +101,14 @@
...
-                  <b>{% fontawesome_icon 'plus-square-o' %} Add topography</b> button.
+                  This surface has no measurements yet.
+                  You can add measurements by pressing the
+                  <b>{% fontawesome_icon 'plus-square-o' %} Add measurement</b> button.
                 </div>
               {% else %}
-                <div class="alert alert-info">
+                <div class="topobank-info-text">
                   This bandwidth plot shows the range of length scales that have been measured for this surface.
+                  Each of the blocks below represent one measurement.
```
mcrot, line 111: "represents"

topobank/analysis/views.py
```diff
@@ -24,13 +24,11 @@
 from bokeh.layouts import row, column, grid
 from bokeh.models import ColumnDataSource, CustomJS, TapTool, Circle, HoverTool
 from bokeh.palettes import Category10
-from bokeh.models.formatters import FuncTickFormatter
```
mcrot, line 27: This breaks the contact plots. It creates an exception when pressing a point in the contact plot:
```python
  File "/opt/conda/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/app/topobank/analysis/views.py", line 1277, in contact_mechanics_data
    'contact-pressure': _contact_mechanics_geometry_figure(
  File "/app/topobank/analysis/views.py", line 1153, in _contact_mechanics_geometry_figure
    colorbar.formatter = FuncTickFormatter(code="return format_exponential(tick);")
NameError: name 'FuncTickFormatter' is not defined
```

topobank/analysis/views.py
```diff
@@ -1293,38 +1273,32 @@ def contact_mechanics_data(request):
...
-                    title="Contact geometry",
                     **geometry_figure_common_args),
                 'contact-pressure': _contact_mechanics_geometry_figure(
                     pressure,
-                    title=r'Pressure',
-                    value_unit='E*',
                     **geometry_figure_common_args),
                 'displacement': _contact_mechanics_geometry_figure(
                     displacement,
-                    title=r'Displacem.', value_unit=unit,
```
mcrot, line 1305: Now "None (µm)" is printed there .. this is not what we want?

mcrot: Same for "Gap", for contact pressure it's "None (None)". For the latter the unit should be "E*" as we discussed in #592 .

mcrot (changes requested):
We need to improve some more things:
- [x] contact plot does not fully work any more due to missing import, try pressing a point
- [ ] colorbar labels and units in the contact plot details are broken (see single comments)
- [ ] there are several more places where "topography" is used and it not replaced by "measurement": Buttons in Analysis plots, table for special values,INFORMATION sheet in Excel files, text files from analysis, landing page of the site, maybe also subject type in task  information (would need to change "label()" method in classes), maybe also in warnings in contact plots
- [ ] we should not include changes of directory `.idea` in this PR, let's decide separately about the Python interpreter used

I would also like to discuss, probably that's easier in person:
- the text over the gallery is more informative now, but in my opinion visually less appealing; can we change this, e.g. with some box, more padding etc? I've liked the alert-info box  before, maybe we can use sth. similar?
- the new contact plot layout looks nicer than before - how important is to compare e.g. the displacement of different loads? This was easier in the version before, because both plots were visible at the same time. Resizing in the browser does not work here, so the user has to scroll up and down.

topobank/analysis/views.py
mcrot: Suggestion: With the new layout the physical quantity is visible on the left, so it would be sufficient to just display
the unit over the colorbar?

topobank/analysis/views.py
mcrot: Fixed.

topobank/templates/manager/surface_detail.html
mcrot: Fixed.

topobank/templates/manager/surface_detail.html
```diff
@@ -48,15 +48,17 @@
 
             <div class="tab-pane fade active show" id="topographies">
               {% if surface.num_topographies == 0 %}
-                <div class="alert alert-info">
-                  This surface has no topographies yet.
-                  You can add topographies by pressing the
-                  <b>{% fontawesome_icon 'plus-square-o' %} Add topography</b> button.
+                <div class="topobank-info-text">
+                  This surface has no measurements yet.
```
mcrot, line 52: In case that there are no topographies yet both the bandwidth plot and the gallery show a little info text.

Now this is displayed in two different ways: For the gallery with class "topobank-info-text" and for bandwidth plot with "alert-info". I think this should be consistent and "alert-info" is a better here, because a user action is required in order to see something. What do you think?

pastewka: Okay, alert-info is fine with me here.

pastewka:
Some "topographies" may be left, we will update as we find them in the code or the interface. Most should have been renamed to "measurements".

@mcrot - Can you have a look at this PR again?

pastewka:
Hang on @mcrot - I still need to update the CSS for the Bokeh tabs

pastewka:
Okay, I've updated the size of the Bokeh tabs and buttons - @mcrot you can have another look at this

topobank/templates/tabnav/tabs.html
```diff
@@ -36,7 +36,7 @@
                href="{% url 'manager:download-selection' %}"
                type="button"
                id="download-selection">
-              <i class="fa fa-download"></i> Download related surfaces
+              <i class="fa fa-download"></i> Download selected surfaces
```
mcrot, line 39: What I have implemented here is not the downloaded of the selected surfaces, but
the download of all surfaces which are related to the selection. That means 
- if a topography is selected, its surface is downloaded
- if a tag is selected, all surfaces are downloaded which have surfaces or topographies with this tag

So think "selected surfaces" is misleading here. Maybe there is a better word?

topobank/templates/analysis/contactmechanics_card_detail.html
```diff
@@ -117,7 +102,7 @@ <h3>{{ title }}</h3>
                            title="Type of calculation"
                            data-content="This option determines how the elastic interactions are calculated.
                                    This affects edge effects that may show up in the results at large contact area.
-                                   Calculations can assume that the surface repeats periodically or that it is pushing down on a nonperiodic,
+                                   Calcula  tions can assume that the surface repeats periodically or that it is pushing down on a nonperiodic,
```
mcrot, line 105: Is this word warp intended? Or a hyphen needed?

mcrot:
There are some more occurences of "topographies" I haven't seen before, at least on sharing info page and help page. I' have a look at that.

pastewka:
> There are some more occurences of "topographies" I haven't seen before, at least on sharing info page and help page. I' have a look at that.

Okay, thanks!

pastewka:
There is also a problem with the tests, I'll have a look at those. It's related to the renaming I think.

mcrot (changes requested):
Fine for me, except of the naming of the new download button. Please see my comment,
I think the old label was more correct in specifying what the button does.

And there is one more comment about a missing hyphen or broken word, I don't know what was intended there.

topobank/templates/tabnav/tabs.html
pastewka: I changed this because "related" is confusing, what is a "related" surface? I understand what you mean but the average user will not. I think selected is good enough. I understand that this includes measurements that my not have been selected, but would not emphasize this in the GUI.

topobank/templates/analysis/contactmechanics_card_detail.html
pastewka: No this is a bug...

topobank/templates/tabnav/tabs.html
pastewka: We could just remove the "selected" -> "Download surfaces"

pastewka:
I've renamed the download thing to just "Download surfaces". I've also updated the tests.

mcrot (approved):

mcrot:
Thank you!

PR #654, opened 2021-05-05
Change RMS value table to show generic roughness parameters

mcrot:
The renaming of the analysis function will be done later: #653 .

Could you have a look:
- How can curvature values be calculated? The method `rms_curvature_from_profile` is not available for uniform line scans and for 2D data.
- When to use "RMS gradient" and "RMS Slope" as quantity names?
- Is the list of values correct and complete? See `functions.py`

### DIFF
--- a/topobank/analysis/functions.py
+++ b/topobank/analysis/functions.py
@@ -1222,35 +1222,34 @@ def rms_values(topography, progress_recorder=None, storage_prefix=None):
     #
     # RMS gradient/slope
     #
+    result.extend([
+        {
+            'quantity': 'RMS slope',
+            'from': FROM_1D,
+            'symbol': 'R&Delta;q',
+            'direction': 'x',
+            'value': topography.rms_slope_from_profile(),  # x direction
+            'unit': 1,
+        }
+    ])
     if is_2D:
         result.extend([
             {
-                'quantity': 'RMS gradient',
-                'from': '',
-                'symbol': 'R&Delta;q',  # HTML
-                'direction': 'x',
-                'value': topography.rms_gradient(),  # x direction
-                'unit': 1,
-            },
-            {
-                'quantity': 'RMS gradient',
-                'from': '',
-                'symbol': 'R&Delta;q',
-                'direction': 'y',
-                'value': topography.transpose().rms_gradient(),  # y direction
-                'unit': 1,
-            },
-        ])
-    elif topography.dim == 1:
-        result.extend([
-            {
-                'quantity': 'RMS slope',  # TODO slope or gradient?
+                'quantity': 'RMS slope',
                 'from': FROM_1D,
-                'symbol': 'R&Delta;q',
-                'direction': 'x',
-                'value': topography.rms_slope_from_profile(),
+                'symbol': 'R&Delta;q',  # HTML
+                'direction': 'y',
+                'value': topography.transpose().rms_slope_from_profile(),  # y direction
                 'unit': 1,
-            }
+            },
+            {
+                'quantity': 'RMS gradient',
+                'from': FROM_2D,
+                'symbol': '',
+                'direction': '',
+                'value': topography.rms_gradient(),
+                'unit': 1,
+            },
         ])
 
     return result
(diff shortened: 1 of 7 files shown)


topobank/analysis/functions.py
```diff
@@ -1139,62 +1139,119 @@ def rms_values(topography, progress_recorder=None, storage_prefix=None):
...
+                'direction': None,
+                'value': topography.rms_height_from_area(),
+                'unit': unit,
+            },
+        ])
+    #
+    # RMS curvature
+    #
+    if is_2D:
+        result.extend([
```
pastewka, line 1193: That's correct, it doesn't exist yet. We should implement this at some point but I don't think it is crucial that we have this now.

topobank/analysis/functions.py
```diff
@@ -1139,62 +1139,119 @@ def rms_values(topography, progress_recorder=None, storage_prefix=None):
...
-        dh_dx, dh_dy = topography.derivative(n=1)
+    #
+    # RMS gradient/slope
+    #
+    if is_2D:
         result.extend([
             {
-                'quantity': 'RMS Slope',
+                'quantity': 'RMS gradient',
+                'from': '',
```
pastewka, line 1229: This should be `FROM_2D`

topobank/analysis/functions.py
```diff
@@ -1139,62 +1139,119 @@ def rms_values(topography, progress_recorder=None, storage_prefix=None):
...
+                'symbol': 'R&Delta;q',  # HTML
                 'direction': 'x',
```
pastewka, line 1231: The gradient has no direction. This also means the next entry needs to be removed.

topobank/analysis/functions.py
```diff
@@ -1139,62 +1139,119 @@ def rms_values(topography, progress_recorder=None, storage_prefix=None):
...
-        dh_dx = topography.derivative(n=1)
         result.extend([
             {
-                'quantity': 'RMS Slope',
+                'quantity': 'RMS slope',  # TODO slope or gradient?
+                'from': FROM_1D,
+                'symbol': 'R&Delta;q',
                 'direction': 'x',
-                'value': rms_slope_from_der(dh_dx),
+                'value': topography.rms_slope_from_profile(),
```
pastewka, line 1251: Looks good, but this exists also for 2D topographies, and there it can be computed in x and y direction. Those two entries should be added to the 2D case.

pastewka (changes requested):

topobank/analysis/functions.py
```diff
@@ -1139,62 +1139,119 @@ def rms_values(topography, progress_recorder=None, storage_prefix=None):
     # Get low level topography from SurfaceTopography model
     topography = topography.topography()
 
+    # noinspection PyBroadException
     try:
         unit = topography.info['unit']
         inverse_unit = '{}⁻¹'.format(unit)
-    except:
+    except Exception:
```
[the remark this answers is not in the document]
mcrot: Yes!

topobank/analysis/functions.py
mcrot: I assume you mean `rms_curvature_from_profile` here for uniform line scans and for 2D.

topobank/analysis/functions.py
mcrot: Okay, thanks.

topobank/analysis/functions.py
mcrot: Added.

pastewka:
Hi @mcrot - the gradient thing is still wrong, I'll edit it locally and push it to this branch

pastewka:
I just pushed this here

mcrot:
I hadn't pushed it yet..

pastewka:
Okay

pastewka:
I was marked as resolved - that's why I thought this was current.

mcrot:
> I was marked as resolved - that's why I thought this was current.

Yes, thanks for checking again! I'm not sure yet about best practices for some PR features in GitHub, e.g. regarding the "Resolve" button or the "Draft" mode.

Now this PR should be complete.

Issue #743, opened 2021-10-20
Avoid scientific notation on linear axis


AntoineSIMTEK:
<img width="224" alt="grafik" src="https://user-images.githubusercontent.com/40853209/138099774-9128ded0-8926-43f9-bd33-1aeb9c3f2ff1.png">

Here the x and y axis use $2x10^2, 3x10^2$ ,  but it would better to use simply 200, 300 for readability.

This comment only applies to linearly scaled axis !

mcrot:
Good point. Would do you think @pastewka?
We could also discuss that in the next meeting.

Issue #410, opened 2019-12-13
Analysis fails for 8k by 8k surface...


pastewka:
...and possibly larger. I uploaded the 8k HDF5 from the contact mechanics challenge. The PSD analysis fails with a memory error, the other analyses just stall. We should either limit the topography size or make this work (the latter preferred). This may require moving the Computational part to a separate server.

pastewka:
I need to check that this has been resolved

pastewka:
This still seems to be an issue. There is also no easy way to retriggered a failed calculation.

mcrot:
Yes, once there was a trigger button (#318 ), I would also like to have something like this again.
For the user, but also in an admin panel with overview over all analyses and filters for their states (running/pending/started/failure).

Hm - we are already on a larger server. Is it still a memory error?

pastewka:
It complains about a redis connection loss. My guess is the process is killed because of memory issues.

pastewka:
We now have a trigger button again for all analyses and for cache renewal.

pastewka:
This is still an issue. DZI creation works but analysis fails with an error message that is not particularly explanatory:

![image](https://github.com/ContactEngineering/topobank/assets/2530391/eb085a70-fa58-47cf-9b51-a755e2a2a612)

pastewka:
This is not because of memory issues, but Redis limits. Disable caches fixes this. Putting the whole topography data into the cache does not make sense, it is bound to be slow as it's pickled.

Issue #340, opened 2019-08-01
Check whether errors during file loading are displayed well


mcrot:
During implementation of #289, topography files will not be fully loaded after upload.
So far, an error was displayed there and we keep this, but an error after upload will
only reflect whether the format could be detected and whether the header of the file is correctly read. So what happens, if the file is broken later? How can we test this?

PR #758, opened 2021-10-26
ENH: OpenSeadragon for map displays & new layout for search results

pastewka:
This is a draft MR for enabling map display through OpenSeadragon. Task breakdown:

- [x] Write DZI topography images to S3
- [x] Render using OpenSeadragon
- [x] Remove DZI from S3 when topography is removed
- [x] Write DZI images to S3 for analysis results that have maps (e.g. contact mechanics)
- [x] Render analysis results using OpenSeadragon
- [x] Remove DZI from S3 when analysis is removed


pastewka:
The views of 2D maps now work as I renvisioned them. Note that the you need a development version of `SurfaceTopography` to run this.

I need a bit of help at this point @mcrot :
1) How do I add a JS library (OpenSeadragon) to the project properly? Is there a package manager that I need to use. I just dumped them in `static/js` for now.
2) Do I need to take care of deleting analysis results or is everything beneath `/media/analyses/XXX` simply removed if the analysis disappears.
3) I am generating a set of files (Deep Zoom Images) for each topography. They are currently dumped into `/media/topographies` (the derived `squeezed` files are also there). Is this the right place or should we have a `/media/derived` for dependent but derived data? How do I implementing deleting these files when the topgography disappears?

pastewka:
Things left to do:
- [] Display of topographies with aspect ratio != 1

pastewka:
I've created an issue for the aspect ratio #764

pastewka:
Hi @mcrot - I think this is ready for your review

pastewka:
Hi @mcrot - I'll be pushing a couple of updates to the tests

pastewka:
This now tests creation and deletion of thumbnails and DZI files

pastewka:
Hi @mcrot - I am reverting this to draft. Will address #769 and #479 here

pastewka:
Hi @mcrot - in order to avoid loading the topography in the Django server, I had to remove the following check:

https://github.com/ContactEngineering/TopoBank/blob/63ab8506decbff914a6ee06efc464e20a8a1776b/topobank/manager/views.py#L439

This `renew_squeezed_datafile` is used to check whether the file is corrupted. However, I don't think we need this check, since the datafile is already opened here:

https://github.com/ContactEngineering/TopoBank/blob/63ab8506decbff914a6ee06efc464e20a8a1776b/topobank/manager/forms.py#L108

It is opened (i.e. metadata is read and parsed), but the data (i.e. height information) is not read. Do you know of a case where opening the file works but reading the data does not? The readers should make a guarantee that if the file can be opened, it will be possible to read data. (I can envision situations where the current readers do not guarantee this, but this would be weirdly malformed files.)

pastewka:
Final todo is to store bandwidth in database.

mcrot:
> It is opened (i.e. metadata is read and parsed), but the data (i.e. height information) is not read. Do you know of a case where opening the file works but reading the data does not? The readers should make a guarantee that if the file can be opened, it will be possible to read data. (I can envision situations where the current readers do not guarantee this, but this would be weirdly malformed files.)

This check was introduced here: #340 .
No, I don't know of such a case except the ones I've created artificially. I think I just wanted to make
sure that there is only valid data in the system.

So for me it's okay to remove to check. Maybe we should make sure that the size of measurements is limited,
so too large measurements will not fill up the memory of the analyses server again and again?! Also see #410.

pastewka:
Okay, I will look at your checks. I think this needs to be handled on the `SurfaceTopography` side. If there is a data buffer, the readers should check whether the file is actually large enough to contain it. Many readers already do that implicitly, but not all.

pastewka:
(Note that since it now runs in Celery, it won't produce an internal server error.)

pastewka:
Hi @mcrot - this PR is ready for your review. It is also addressing the issue of avoiding to load the topography within the Django server. (It still does that for line scans, but that is okay.)

mcrot:
Great, thanks! I'll have a look.

mcrot (approved):
Looks good. I've added the PNG download buttons for contact mechanics.
I've tried to show another indicator because sometimes it takes some more 
seconds on my laptop to show the map when the view has already loaded, but
I haven't found a suitable event in OpenSeaDragon - maybe later.

I'll try to fix the basket and menu position in another issue in order to have this PR finished.

pastewka:
Closes #743

Issue #759, opened 2021-10-28
Messy rendering of input in contact analysis


mcrot:
Here the input for "Pressures" seems to have a too small height:

![grafik](https://user-images.githubusercontent.com/9113541/139228423-7b53d208-7c90-484c-9633-a39f36dd2b55.png)

pastewka:
This is already fixed.

Issue #725, opened 2021-09-22
Cards in analysis view scale weirdly when buttons are clicked


pastewka:
I clicked the "Average / Measurement" button:

![grafik](https://user-images.githubusercontent.com/2530391/134356273-bde7284e-2915-467e-bb1c-4f4a75886402.png)

The behavior depends on how large the whole window is. Maybe we need to break the labels in multiple lines if they are too long.

pastewka:
Also, the button could span the whole width of the card:

![grafik](https://user-images.githubusercontent.com/2530391/134356037-a227c899-9648-4ac4-a08f-9d24819d68c4.png)

PR #775, opened 2021-12-15
MAINT: Tweaks to plotting and visualization

pastewka:


pastewka:
Things left to do:

- [ ] Improve choice of unit
- [ ] Switch contact mechanics to Vue Bokeh component
- [ ] Fix tests

pastewka:
Hi @mcrot - this is ready for your review

pastewka:
Things left to do:
- [] Replace AXIOS by fetch API
- [] Code that handles promises should use async/await syntax
- [] Think about whether Webpack bundler makes sense
- [] Think about whether using TypeScript makes sense (requires Webpack bundler)

pastewka:
This PR already fixed #759

pastewka:
This PR will fix #725

pastewka:
This PR addresses #792

pastewka:
Maybe ˋget_valid_nameˋ does exactly the normalization of the name.

topobank/analysis/functions.py
```diff
@@ -358,13 +369,12 @@ def curvature_distribution(topography, bins=None, wfac=5, progress_recorder=None
     #
     if topography.dim == 2:
         curv_x, curv_y = topography.derivative(n=2)
-        curv = curv_x + curv_y
+        curv = (curv_x + curv_y) / 2
```
mcrot, line 372: Does this mean, we have to recalculate all curvature distributions?

mcrot (changes requested):
I've still got some problems when using this branch my data.
Please have look at my questions regarding specific lines of code.

topobank/analysis/views.py
```diff
@@ -618,7 +585,7 @@ def get_axis_type(key):
                 #
                 # Actually plot the line
                 #
-                show_symbols = np.count_nonzero(~mask) <= MAX_NUM_POINTS_FOR_SYMBOLS
+                show_symbols = s['x'] <= MAX_NUM_POINTS_FOR_SYMBOLS
```
mcrot, line 588: Is here some length operator missing, because s['x'] is an array in general?
I've replaced here `s['x']` with `s['x'].size` which worked for many cases.

Then suddenly I've had an error again if the series has only one value. Then `s['x']` is an int.
I think this is due to the JSON serializer and because an array with a single value is serialized as int.
It would be best to always have an array here in JSON, also for one value, because otherwise we always need to handle it.

Is there a way to always serialize as array?

topobank/analysis/functions.py
pastewka: Yes

pastewka:
Things left todo:
- [] Fix `results_obj`, read from S3 and automatically piece together from multiple files
- [] Remove redundant pickled `result` dictionary from database
- [] Add dash to source indicator

mcrot:
I've just seen this exception while testing:
```python
django_1        | Traceback (most recent call last):
django_1        |   File "/app/manage.py", line 30, in <module>
django_1        |     execute_from_command_line(sys.argv)
django_1        |   File "/opt/conda/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
django_1        |     utility.execute()
django_1        |   File "/opt/conda/lib/python3.8/site-packages/django/core/management/__init__.py", line 395, in execute
django_1        |     django.setup()
django_1        |   File "/opt/conda/lib/python3.8/site-packages/django/__init__.py", line 24, in setup
django_1        |     apps.populate(settings.INSTALLED_APPS)
django_1        |   File "/opt/conda/lib/python3.8/site-packages/django/apps/registry.py", line 114, in populate
django_1        |     app_config.import_models()
django_1        |   File "/opt/conda/lib/python3.8/site-packages/django/apps/config.py", line 301, in import_models
django_1        |     self.models_module = import_module(models_module_name)
django_1        |   File "/opt/conda/lib/python3.8/importlib/__init__.py", line 127, in import_module
django_1        |     return _bootstrap._gcd_import(name[level:], package, level)
django_1        |   File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
django_1        |   File "<frozen importlib._bootstrap>", line 991, in _find_and_load
django_1        |   File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
django_1        |   File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
django_1        |   File "<frozen importlib._bootstrap_external>", line 783, in exec_module
django_1        |   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
django_1        |   File "/app/topobank/manager/models.py", line 29, in <module>
django_1        |     from bokeh.models import DataRange1d, LinearColorMapper, ColorBar, FuncTickFormatter
django_1        | ImportError: cannot import name 'FuncTickFormatter' from 'bokeh.models' (/opt/conda/lib/python3.8/site-packages/bokeh/models/__init__.py)
```
`bokeh` version 2.4 has this import, but maybe not version 3 of bokeh? 
Since it is not officially released yet, do we need it?

mcrot:
Ah I see, there was already a commit because of `FuncTickFormatter` (https://github.com/ContactEngineering/TopoBank/pull/775/commits/2caa1f28f86c1c84e70ee08881cc43d81405b9f5), maybe this has been overwritten?

mcrot:
I've worked a bit with on branch and I think we shouldn't merge it yet. 
There are still some issues, of which some need a bit more time:

- Power Spectrum: Selection of data series does not match the data shown, e.g. "reliable" data is not a subset of the data when chosen the "unreliable" option
- Legends: Always all measurements are shown, also those which are not selected; series names are not included in the legend, only the measurement name; dashes are not used yet to distinguish series which all look the same 
- Celery progress bars cause Vue errors
- There are new color palettes for PlotCardView - let's check whether the contact plot already uses the same

pastewka:
Okay sounds good. Fix this before merging.

pastewka:
Does this mean moving away from `celery-progress`? I'd be okay with this as the Celery API leaks too much through. We should try to move towards clean(er) API for the analysis part of the code.

mcrot:
@pastewka , could you have a look at the contact calculation at functions.py in this branch?
Especially here where now `assign_patch_numbers` and `patch_areas` are undefined: https://github.com/ContactEngineering/TopoBank/blob/42b234e65fedadd37acaa41f407003d752a102f2/topobank/analysis/functions.py#L1288
After merging develop I'm currently not sure there is just an import missing or this has to be changed completely.

pastewka:
This was a missing import. I just pushed a patch, let's see if CI passes.

mcrot:
Yes, we're back on green. I remove the draft status.

pastewka:
Awesome!

mcrot (approved):

Issue #801, opened 2022-02-25
Store raw data under a unique prefix


pastewka:
Raw files are currently uploaded to the same prefix. This means Django attaches suffixes to avoid file name collisions. We should just store the raw file under the prefix that we currently use for derived data and use the original file name.

This leads to problems when downloading containers, as all files are in the toplevel ZIP folder and this could lead to name collisions. We should probably also introduce a (consecutively numbered) subdirectory for each measurement in the container.

Any thoughts on this @mcrot ?

mcrot:
Currently there is already some code which prevents name collisions for measurements in the ZIP container. 
If  a filename is already used in the ZIP, a counter is added to the file name.

Probably this could be solved in a better way. We could use subdirectories which name is a number. A natural choice would be just the measurement id in production as this is already used in the URL. If we don't want to
further expose these numbers, we could also use a consective number, e.g. over all measurements for a surface or over all measurements in the ZIP.

I'm not sure what you are meaning here with derived data - the analyses?

>  We should just store the raw file under the prefix that we currently use for derived data and use the original file name.

The prefix or subdirectory we use for measurement files are currently named with the user id.

pastewka:
We don't need the user prefix, in particular since we want to implement transfer of ownership at some point. See #706

pastewka:
Files in the ZIP container should be prefixed by a unique consecutive topography id, e.g. `001_random.txt`, to avoid name collisions. If there is more than one digital surface twin in the container, then add an additional twin index, e.g. `01_001_random.txt`.

Issue #787, opened 2022-01-24
Refactor thumbnail generation without selenium


mcrot:
At the moment (version 0.18), thumbnails are generated by plotting with bokeh and saving a picture using selenium.
This feels a bit clumsy with timeouts etc. We should replace that e.g. by generating images with matplotlib are
taking the top level result from the DZI files.

Issue #307, opened 2019-07-18
URL should be designed such that a certain view can be bookmarked


pastewka:
It would be useful to send certain views to colleagues or to bookmark them. The URL should represent the state of what is seen on the screen.

pastewka:
This has been implemented.

PR #802, opened 2022-03-01
MAINT: Remove Selenium dependency and harmonize storage locations

pastewka:
This PR will address:
- Selenium dependency, #787 
- Storage locations, #801


pastewka:
@mcrot - I've run into a problem and need your advise here:

Both `Topography` and `Analysis` models now have a `storage_prefix` attribute that defines the prefix in the storage system. It need to know the `id` of the model to generate the prefix. The `id` is only known once the model has been saved to the database. This cause problems in the tests, where the model is initialized with some data file and at this point the `id` is not yet known. The solutions to this that I've found are:

- Switch the `id` field to a UUID (see here: https://stackoverflow.com/questions/41428290/django-instance-id-none-when-uploading-image).
- Overload the `save` method (see here: https://stackoverflow.com/questions/9968532/django-admin-file-upload-with-current-model-id)

I'd actually be in favor of using UUIDs for primary keys as this may have security advantages (you cannot guess an id). It may also help here: #307 . However switching to UUIDs would require a larger migration of the database.

pastewka:
I've implemented option 2 and that seems to work

pastewka:
Hi @mcrot - this is ready for your review

topobank/manager/signals.py
```diff
@@ -47,29 +47,46 @@ def delete_datafile(datafile_attr_name):
...
+        for filename in filenames:
+            _log.info(f'Deleting file {path}/{filename}...')
+            default_storage.delete(f'{path}/{filename}')
+        for directory in directories:
+            _log.info(f'Deleting directory {path}/{directory}...')
+            delete_directory(f'{path}/{directory}')
+            default_storage.delete(f'{path}/{directory}')
+
+    datafile_path = instance.datafile.name
+    squeezed_datafile_path = instance.datafile.name
```
mcrot, line 62: Should this be `instance.squeezed_datafile.name`?

topobank/manager/signals.py
```diff
@@ -47,29 +47,46 @@ def delete_datafile(datafile_attr_name):
...
+    # Otherwise we abort deletion.
+    if datafile_path is not None and not datafile_path.startswith(instance.storage_prefix):
+        _log.warning(f'Datafile is stored at location {datafile_path}, but storage prefix is '
+                     f'{instance.storage_prefix}. I will not attempt to deleting everything at this prefix.')
+        return
+    if squeezed_datafile_path is not None and not squeezed_datafile_path.startswith(instance.storage_prefix):
+        _log.warning(f'Squeezed datafile is stored at location {squeezed_datafile_path}, but storage prefix is '
+                     f'{instance.storage_prefix}. I will not attempt to deleting everything at this prefix.')
+        return
+    if datafile_path is not None and not datafile_path.startswith(instance.storage_prefix):
```
mcrot, line 81: Why has this been doubled?

mcrot (changes requested):
Please review my questions on `signals.py`.

topobank/manager/signals.py
pastewka: Can be removed

mcrot (approved):

PR #840, opened 2022-05-31
22 unreliable bw 825 roettger

yaelgail:
Could you please look at these changes ? Does the unreliable bandwidth looks good when it's very little ?


pastewka:
Code looks good, I have not run it yet will do so today or tomorrow.

pastewka (changes requested):
1) If all bandwidth is reliable, there is still a little red box to the left. It may make sense to keep the cutoff as "None" to indicate that all data is realiable and not plot that box.

2) You need to recompute the short cutoff when the user changes the tip radius (or edits the instrument)

3) I would separate the creation of the squeezed data file from updating the bandwidth information. One reason is that now the computation of the reliability cutoff is probably the most time consuming part (and no longer the creation of the squeezed file). Also, you don't need to recreate the squeezed file when you edit the instrument.

Can you implement these 3 changes @yaelgail ?

pastewka:
Here is a screenshot showing the red boxes for fully reliable measurements:

image

pastewka:
I would also change the title here to "Measurement artifacts"

image

pastewka:
PS: Can you also update the changelog?

mcrot (approved):

mcrot:
> 1. If all bandwidth is reliable, there is still a little red box to the left. It may make sense to keep the cutoff as "None" to indicate that all data is realiable and not plot that box.
> 
>     2. You need to recompute the short cutoff when the user changes the tip radius (or edits the instrument)
> 
>     3. I would separate the creation of the squeezed data file from updating the bandwidth information. One reason is that now the computation of the reliability cutoff is probably the most time consuming part (and no longer the creation of the squeezed file). Also, you don't need to recreate the squeezed file when you edit the instrument.
> 
> 
> Can you implement these 3 changes @yaelgail ?

I've just implemented the third point.

mcrot (approved):

PR #849, opened 2022-06-27
Switch to SurfaceTopography v1.0

pastewka:
This PR implements API changes from latest SurfaceTopography


topobank/manager/views.py
```diff
@@ -312,9 +312,7 @@ def clean_channel_name(s):
             # Set data source choices based on file contents
             #
             kwargs['data_source_choices'] = [(k, clean_channel_name(channel_info.name)) for k, channel_info in
-                                             enumerate(channel_infos)
```
pastewka, line 315: Hi @mcrot - please have a look at this change. I did not undertstand what the condition was good for. Unit information in the `info` dictionary is now gone.

mcrot (dismissed):
Looks good!

mcrot:
I'm not sure which PR should be merged as first. Probably this one?

pastewka:
This is because I think upgrading the dependencies may lead to unforseen issues.

mcrot (approved):

Issue #847, opened 2022-06-22
Web app is slow


tdbjacobs:
Ever since the update, the web app is running extremely slow. 
Not only do analyses move slowly, but if you refresh the screen after an analysis, it seems to recalculate that -- rather than just showing the previously computed analysis
-Tevis

pastewka:
It does not recalculate. It is just slow at loading things in the background. I am not sure what it is.

This has to be the first priority of @mcrot when he comes back from vacation.

tdbjacobs:
@pastewka, a naive question: 

I know the web app stores the results of prior analyses so that it doesn't need to recompute. The purpose of this is to reduce waiting time if a user performs an analysis that has already been performed. 

However, it appears that the biggest delay is in visualization, not in calculation.  So is it possible to store the visualization of the prior analysis also? 

Probably this is not feasible, but if it was, this would help to achieve the same objective -- fast loading of a recomputed analysis.

mcrot:
@tdbjacobs , we think the performance improvements were effective. Could you have a second look how fast it is now on your side?

Issue #841, opened 2022-06-03
Loading of graphs extremely slow


pastewka:
One of the main issue we wanted to fix with PR https://github.com/ContactEngineering/TopoBank/pull/775 was the slow loading of large digital surface twins. Unfortunately, the loading is *still* very slow. Loading the data is actually fast, but it takes a long time until the Vue app that initialized the Bokeh plot seems to be ready. Bokeh is then nicely loading the data in the background and it appears as it becomes available. I am not sure what exactly is going.

This issue did not occur in the past when developing locally on my computer.

pastewka:
It is even worse than that, I now get this error:

![image](https://user-images.githubusercontent.com/2530391/171814299-13034c8e-6706-4ccf-a61f-b802fb018925.png)

but no error mail reporting details.

pastewka:
The error went away after I recomputed the analyses. We should probably have more sensible error messages and a button to recompute.

pastewka:
I have a suspicion about what is going on. `PlotCardView` generates a list of all data sources, including their URL. On S3 this means creating a presigned URL. We have no control over how slow that is - this may be the bottleneck as it requires a request to the S3.

Solution: Just pass the (unsigned) data URL, we have an automatic redirect that takes cares of signing if the user is logged in.

pastewka:
PR #848 implements avoiding the presigning but it is now actually slower in developement - the problem is that it is difficult to predict what will happen in production.

PR #855, opened 2022-07-07
Performance tweaks

pastewka:
Performance tweaks addressing issues #841 and #847


pastewka:
This is useful reading: https://www.mattduck.com/2021-01-django-orm-result-cache.html

PR #856, opened 2022-07-07
Performance tweaks

pastewka:
Addresses performance issues with analysis - #841 #847


pastewka:
Hi @mcrot - here are the performance tweaks. They should get rid of most of the bottlenecks.

pastewka:
Hi @mcrot - I've removed package updates from this PR. Note that some analysis results will need to be recomputed. (Those that use `PlotCardView`.)

config/settings/local.py
```diff
@@ -61,9 +61,9 @@
 # django-debug-toolbar
 # ------------------------------------------------------------------------------
 # https://django-debug-toolbar.readthedocs.io/en/latest/installation.html#prerequisites
-INSTALLED_APPS += ['debug_toolbar']  # noqa F405
```
mcrot, line 64: Is it intended that we disable the debug tool for development by default?
Or is this just a left-over from debugging?

mcrot (changes requested):
Looks great! I'm looking forward to see this in production.

Two things:
- Shall we really disable the debug toolbar for development by default?
- The test for split dictionaries should be extended by the use of the supplementary dict.

mcrot (approved):
This can be merged. If there are no string reasons against, I would keep the debug toolbar.

config/settings/local.py
pastewka: I am readding it

mcrot (approved):

PR #861, opened 2022-07-22
BUG: Wrong contact mechanics details

### LINKED ISSUE #859: Incorrect understanding/bug displaying analysis results?
My focus is on the study of evolution of the contact geometry and gap of a randomly rough surface as a function of the applied pressure, and I have some weird results that I don't understand. I am a new user, so I might be wrong in my understanding - feel free to correct me.

The fractional contact area (A/A0) shows an expected trend - increasing with increasing pressure. If my understanding is correct, one would expect the contact patches (white regions) to increase and gaps (bright green regions) to reduce, as the fractional contact area increases. However, the contact geometry and gaps do not show this in the result image on the right-hand side when I click on the individual data points, as you can see in the image attached (the five images are corresponding to the first five datapoints, in order, left to right). 
image (1)

N.B. The scale bar is same in all the images, so it is not rescaling to a smaller range, and I have also changed the max # of iterations to 500, in case it is a convergence issue.

Is there something wrong in my fundamental understanding, or is this a bug where the images being shown are different from the data points being clicked on?

pastewka:
Clicking on a data point in the contact mechanics view showed the wrong details at the selected pressure, This closes #859


topobank/analysis/functions.py
```diff
@@ -1253,7 +1253,7 @@ def contact_mechanics(topography, substrate_str="nonperiodic", hardness=None, ns
         if hardness:
             dataset.attrs['hardness'] = hardness  # TODO how to save hardness=None? Not possible in netCDF
 
-        storage_path = f'{storage_prefix}/step-{i}'
+        storage_path = f'step-{i}'
```
mcrot, line 1256: Why has the `storage_path` been removed here? It seems that this doesn't work. It generates files in the top level
of the storage and the contact maps cannot be found. Could we revert this?

mcrot (changes requested):
The retrieval of contact data does not work any more.
I think the storage path is wrong.

config/settings/local.py
```diff
@@ -61,9 +61,9 @@
 # django-debug-toolbar
 # ------------------------------------------------------------------------------
 # https://django-debug-toolbar.readthedocs.io/en/latest/installation.html#prerequisites
-INSTALLED_APPS += ['debug_toolbar']  # noqa F405
+#INSTALLED_APPS += ['debug_toolbar']  # noqa F405
 # https://django-debug-toolbar.readthedocs.io/en/latest/installation.html#middleware
-MIDDLEWARE += ['debug_toolbar.middleware.DebugToolbarMiddleware']  # noqa F405
+#MIDDLEWARE += ['debug_toolbar.middleware.DebugToolbarMiddleware']  # noqa F405
```
mcrot, line 66: Should this be removed in general? In one of the previous PR, we kept this.

topobank/analysis/functions.py
mcrot: When we remove the storage path here and make it again by changing the code at another place, we would have to recalculate all contact analyses. So I would prefer keeping the storage path here if there aren't strong reasons.

config/settings/local.py
pastewka: Sorry, please keep this.

topobank/analysis/functions.py
pastewka: Okay, keep it here.

pastewka: (The reason it has been removed is that the `storage_prefix` is the prefix in the S3 but it has been used as the prefix of the API. This is of cource wrong. For the API I am now removing everything before the last slash and reconstructing the path.)

pastewka: (But I agree that you are right to keep as it was before.)

mcrot (approved):

mcrot (approved):

PR #918, opened 2022-11-07
MAINT: Remove scipy dependency

pastewka:
This PR bumps NuMPI to 0.3.1 which has no longer a scipy dependency.


pastewka:
We need currently use `scipy` to load Matlab files

pastewka:
This is correct.

mcrot (approved):

pastewka:
We can't get rid of scipy because `SurfaceTopography` depends on the sparse solver for filling missing data points.

PR #921, opened 2022-11-08
MAINT: Added django-watchman

pastewka:
django-watchman allows introspection of whether the endpoint is alive. This is useful to check the status of the app or to integrate it with a load balancer.


pastewka:
@mcrot - do we need a test for this?

mcrot:
I've added an additional check whether a celery worker is available.
For this check I've added test.

mcrot:
For production I've added the need for being logged-in as staff member in order to request the watchman urls,
in order not to make it totally public. I've also added links to the user menu in the application that you can access
the watchman dashboard directly.

In future maybe we want a token authentication instead of the user authentication. It depends who or which process is looking for the status.

mcrot (approved):

pastewka:
I have implemented a simple native health check

mcrot:
Okay .. I think what is good about the celery worker check, is that we can identify connection losses to the workers.

pastewka:
Agreed. Reopening this.

pastewka:
Sorry for the mess, I didn't realize you had worked on this. Watchman is probably better in the long run as it offer more complete health information.

PR #925, opened 2022-11-20
Suggestion for running local development with plugins

pastewka:
@mcrot - here is a suggestion on how to include plugins in the local development process. You can edit plugings and/or main TopoBank and Django will dynamically releoad. (Celery won't, but it didn't do that before either.)

It would probably be best if we could maintain a dynamic list of plugins somewhere.


pastewka:
Closing this - superceded by https://github.com/ContactEngineering/topobank-stack-development

PR #927, opened 2022-11-24
MAINT: Bumped SurfaceTopography version to 1.2.2

pastewka:
Updated SurfaceTopography version to 1.2.1, which introduces a Park Systems TIFF reader and crucial bug fixes for VK3 and X3P readers.

PR #928, opened 2022-11-27
MAINT: More robust interpretation of `acquisition_time` info dictionary key

pastewka:
This is required for X3P dates to show correctly in the GUI.


pastewka:
Should go on top of #927

PR #930, opened 2022-11-30
BUG: Fixed datafile location during upload

### LINKED ISSUE #929: Wrong S3 storage location for raw data files
File should be stored under the `raw` prefix of the topograph, but are stored in the user directory.

#801 should have addressed this,.but apparently did not.

pastewka:
Closes #929


mcrot (approved):

mcrot (changes requested):
Overall the code looks correct, but there is a problem in the tests: The "name" attribute is no longer filled correctly. Maybe we have to change the creation of test instances
to use the _upload_path_for_datafile function?! Not sure yet.

mcrot:
Let me have a look at this when the "NaN" problem is solved.

pastewka:
I'll look at the name issue, it would be good have the NaN issue resolved though.

pastewka:
I think there was a missing `instance.save()`

pastewka:
The problem is gone, only failure left is related to the NaN issue.

mcrot:
Great, I'm looking at the NaN issue, I don't have a general solution yet. On Python level, on having `NaN` in JSON is simply ignored and the Parser simply interprets `NaN` (literal, not string) correctly. But when passing it to JS, we need `"NaN"`. I thought I have a solution already, but it only works for `NaN` in numpy arrays to far.

mcrot:
We can delete this PR, correct?

pastewka:
Yes

PR #994, opened 2023-05-30
23 meaningful name surface container download

Siddhantsaka:


topobank/manager/views.py
```diff
@@ -7,7 +7,7 @@
 import dateutil.parser
 import django_tables2 as tables
 import numpy as np
-
+from slugify import slugify
```
pastewka, line 10: Please don't add additional unnecessary dependencies. There is a `slugify` function in Django: `from django.utils.text import slugify`.

topobank/manager/views.py
```diff
@@ -1649,7 +1649,7 @@ def download_surface(request, surface_id):
         except Exception:  # not interested here, why it fails
             renew_publication_container = True
     else:
-        container_filename = DEFAULT_CONTAINER_FILENAME
+        container_filename= slugify(surface.name) #container_filename = DEFAULT_CONTAINER_FILENAME
```
pastewka, line 1652: Please add a space before the equal sign. You also need to add a `.zip` extension to the file name. Please also removed the commented section.

pastewka (changes requested):
As a general comment: Please also fix the failing test. It is related to the naming change.

pastewka (approved):

Issue #967, opened 2023-02-17
web app is really slow and not loading


Cyrian-24:
Waited for 15 minutes and teh website is not loading or computing measurements, thank you!

pastewka:
The web app seems to lock up at certain times. It then takes it 1-2 minutes to recover. I think what is happening is that DJango can handle only a finite number of requests. If *all* of these requests take a long time, then the system locks up until the first finished. At this point, I have no idea what is causing requests to take a long time. My best guess is checking Celery task statuses, i.e. a problem with the message queue.

pastewka:
I think the culprit is the `request_analysis`, as this problem appear to occur when many new analyses are requested.

pastewka:
Hi @Cyrian-24 - we pushed a couple of performance tweaks. Can you give us feedback if the site feels faster now?

pastewka:
Should be partially resolved with #1012 - need to observe what will happen but closing this for now.

PR #1013, opened 2023-08-30
MAINT: Refactoring triggering of cache refresh, analysis into models; also some performance optimizations

pastewka:
This PR reduces the number of SQL request issued in particular when loading analyses. The goal is to reduce load times for large surface twins that can be around 30 seconds.

The strategy is to avoid nested serialization, i.e. return ids of nested datasets and have the frontend load if required. In particular for datasets with many measurements this can be a bottleneck, as each measurement incurs an SQL request to load the complete measurement model, but most of that data is actually not needed by the frontend.

Ideally, the number of database hits for each API call is one. This is not achievable as each request must obtain the user model and query permissions, but below 10 should be possible. (Currently it can beyond 1000 request for DSTs containing 50 or more measurements. This appears to be the key bottleneck making the site slow.)

- [x] #967 
- [x] #1014  
- [x] #1015  
- [x] Check that plugins still work
- [x] Analyses appear to no longer automatically start when uploading a new topography
- [x] #1017 
- [x] Fix tests
- [x] Test migration on production database


pastewka:
We should also try to avoid generic relation, I've opened an issue here #1014

Issue #750, opened 2021-10-23
Move dedicated published and shared tab to filter


pastewka:
![image](https://user-images.githubusercontent.com/2530391/138547347-cf86f265-b025-42f2-b9b8-247d002b99c4.png)

Right now, we have the following filters:
* All accessible surfaces
* Only own surfaces
* Only surfaces shared with you
* Only surfaces published by anyone

My suggestion to add
* Only surfaces shared by you
* Only surfaces published by you

but remove these two dedicated tabs:

![image](https://user-images.githubusercontent.com/2530391/138547383-f5ac06dd-b890-4857-aa79-8a28de6bd5f2.png)

The reason is that these tabs cannot be searched and I now have to many surfaces that are shared with me, that I am not able to find anything in there. The information on who shared it etc. should then be migrated to the "Find & select" pane.

Any thoughts on this @mcrot @tdbjacobs

pastewka:
Related to #982

pastewka:
The filter choices are passed to a Django template here:

https://github.com/ContactEngineering/topobank/blob/fa1b3b3900374574f814f185f65c420a1843385b/topobank/manager/views.py#L797

pastewka:
The template file is called `select.html`

pastewka:
It's actually `SHARING_STATUS_FILTER_CHOICES` not `CATEGORY_FILTER_CHOICES`, but same file.

pastewka:
The actual filtering is done here:

https://github.com/ContactEngineering/topobank/blob/fa1b3b3900374574f814f185f65c420a1843385b/topobank/manager/utils.py#L206

PR #1029, opened 2023-11-08
23 select filter enhancement

IoannisNezis:
This updates the filter in the digital twin select.
New filter options:
- only surfaces shared by you
- only surfaces published by you


topobank/manager/utils.py
```diff
@@ -237,13 +237,26 @@ def filtered_surfaces(request):
...
+                                       .filter(content_type__app_label='manager', content_type__model='surface')
+                                       .exclude(user=user))  # only surfaces
+            surface_ids = [x[0] for x in viewable_surfaces_perms.values_list("object_pk")]
+            qs = qs.filter(creator=user, id__in=surface_ids)  # own surfaces, shared with others
+        case 'published_egress':
+            qs = qs.filter(publication__isnull=False, creator=user)
+        case 'all':
+            pass
+        case other:
+            _log.error(f"Invalid sharing filter: {other}")
```
pastewka, line 259: Should this raise an exception?

IoannisNezis:
Did i miss something in the "git etiquette"?
And should we discuss the content of this merge request here or in the corresponding   Issue #750?

pastewka (changes requested):
Looks good, just one comment

pastewka:
Looks good. I would discuss the content here.

pastewka:
There are some tests that fail, please look at them and try to fix.

topobank/manager/utils.py
IoannisNezis: Yes, i wanted to ask which error would be appropriate.
This error can only occur if the API is used improperly.
So a ValueError?

pastewka: I think a `ValueError` is fine. If this happens, then I think it may trigger an internal server error. However, it should the role of the view to translate that into a proper error message.

IoannisNezis: Actually this code is unreachable
https://github.com/ContactEngineering/topobank/blob/de20c93ef02c4b41d5f013c003757ac65c81a038/topobank/manager/utils.py#L1008-L1027
I'll just remove this case.
Or should i still raise a Error since it would not hurt and maybe increase readability?

IoannisNezis:
Ich würde noch gerne einen test hinzufügen

IoannisNezis:
I notices something:
publishing creates a copy of a surface right?
Is it possible that this copy is shared?
Because i thing this would not be possible to model with the current schema.
Since sharing only sets a permission, a permission published surfaces have as well.

IoannisNezis:
I got no clue why "skip_postgeneration_save" broke. Any ideas?

pastewka:
> I noticed something: publishing creates a copy of a surface right? Is it possible that this copy is shared? Because i thing this would not be possible to model with the current schema. Since sharing only sets a permission, a permission published surfaces have as well.

Publishing creates a copy, but with frozen in permissions. The whole world has view permission, nobody has any of the other permissions. Resharing a published dataset does not make sense as it is visible to everyone anyway.

pastewka:
These trackstat errors can be fixed by adding the `handle_usage_statistics` fixture to the test. (Just add `handle_usage_statistics` to the list of arguments of the test. That argument is never used in the test itself.)

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.4.0 (not yet released)
 
+- ENH: Added select filter options: '(shared|published) by you'
```
pastewka, line 5: Can you tag the issue number (#750) here?

pastewka (changes requested):
One remaining request, the rest looks good

CHANGELOG.md
IoannisNezis: Done

pastewka (approved):

pastewka (approved):

PR #1038, opened 2023-11-14
MAINT: removing DATABASE_URL env var from entrypoint.sh

IoannisNezis:


IoannisNezis:
If `DATABASE_URL` is set for some reason, it will be used.
I found some usages of for example here:
https://github.com/ContactEngineering/topobank/blob/c4a21f7983c9859e004aa3c97c268eabd65828a5/.github/workflows/test.yml#L40-L45

and thought it would be the most reasonable thing to do

pastewka (dismissed):

IoannisNezis:
Question to 'Git etiquette':
Am I supposed to merge, close pull requests and delete feature branches?
Or is that the job of the "product owner"?

pastewka:
You should do the final merge, you are the one that knows best when this thing is ready.

On some projects we set it up that you cannot do it, but I think this is not the case here.

Issue #1016, opened 2023-09-04
Switch Django guardian to direct foreign keys


pastewka:
https://django-guardian.readthedocs.io/en/stable/userguide/performance.html

Note that this change requires critical and tedious testing as bugs in migration could reveals user data.

PR #1044, opened 2023-11-25
ENH: direct foreign keys in django guardian

IoannisNezis:
# direct foreign keys in django guardian
When this pull request is merged the guardian permission model will automatically migrate from using generic foreign keys to real "direct" foreign keys.

## ToDo
- [x] create new model 'SurfaceUserObjectPermission'
- [x] create migration routine
~~- [ ] automatically invoke migration routine at startup~~
- [x] check for permission integrity after migration

At the moment the new model is disabled `enabled = False` and the migration could be tested in production without consequences in case of failure.


IoannisNezis:
@pastewka In issue #1016  you said bugs in migration 'could reveal user data'.
Can you elaborate on that?
Do you mean that if the permissions are migrated incorrectly users could access Datasets they are not supposed to access?
Or is there another Problem, that i'm not aware of?

pastewka:
Hi @IoannisNezis - thanks for taking care of this.

Two comments:
* Regarding your question, data access upon incorrect migration is exactly what I mean. Note that the new direct keys will have a database constraint which the current table does not have. However, I believe there are no dangling permission entries (i.e. cleanup without the constraint has worked okay)
* Regarding the third point on your todo list, migrations run automatically in both development and production stacks so there is nothing that you need to do on your end (except for making sure the table is migrated correctly)

IoannisNezis:
I was a bit imprecise there.
With migration i did not mean the usual database migration, but the migration of the content i.e. the permissions.
My plan is to use a management command, and call it in the entrypoint, like the normal migration.

We spoke about this briefly and i understood that the migration of the database content needs to happen fully automatic.
Is that correct?

pastewka:
Don't use a management command. For an example on how to do a (complex) migration, see here:

https://github.com/ContactEngineering/topobank/blob/main/topobank/analysis/migrations/0018_auto_20230901_2303.py

You have to write a forward and possible reverse function and run it through `RunPython` after you tables are added to the database. This means this would go into the migration file you created and that is already part of this pull request.

IoannisNezis:
I put the migration in the migration file as requested.
I also implemented a management command to check the integrity of the permission migration, in case of paranoia.
The filter: 'shared_egress' was also updated to work with the new permission model.
Note that this filter option will fail as long as the new Permission Model is not enabled.

That's also why the test fails,
how should we handle this?
- enable new Permission Model?
- update filter in separate PR?
- ignore failing test?

topobank/manager/models.py
```diff
@@ -555,6 +556,16 @@ def related_surfaces(self):
         return [self]
 
 
+class SurfaceUserObjectPermission(UserObjectPermissionBase):
+    content_object = models.ForeignKey(Surface, on_delete=models.CASCADE)
+    enabled = False
```
pastewka, line 561: Please simply enable

pastewka (changes requested):

topobank/manager/models.py
IoannisNezis: done

pastewka:
This looks good @IoannisNezis - but there seems to be one remaining issue in one of the tests

IoannisNezis:
Yes, the shared_egress filter included publicized datastets.
Should be fixed now.

IoannisNezis:
I added a `MAINT` instead of a `ENH` to the CHANGELOG, since it does not introduce a new feature but improves a existing feature.
Is that a correct usage of change types?

Issue #904, opened 2022-10-06
Properties for digital surface twins


pastewka:
Digital surface twins should have properties. Those properties can be:

* Categorical
* Numerical

Those should be key value pairs. Categorical properties have discrete values. Numerical properties have continuous values. Numerical properties should have units. (Those should be a free text field initially, we may interpret the units later.)

Categorical properties should *not* be the same as the tags we have currently. Tags are for internal organization (like a folder).

pastewka:
There is currently a tab called "properties" that should be renamed to "description" here:

<img width="260" alt="image" src="https://github.com/ContactEngineering/topobank/assets/2530391/f2bf613f-3dc3-442a-b558-072c702dd2c8">

The idea would be to add a tab with actual "Properties" that may be an editable table.

A "property" has 
1) a name
2) a value (typically a number, but could also be categorical e.g. "good" or "bad", or "blue" or "green")
3) a unit (e.g. "nm", "V", "1", "color")

All of these should be free, but the user should be guided when selecting a name or unit.

pastewka:
As discussed today:

* Here is an example of the validation of data in models: https://github.com/ContactEngineering/topobank/blob/7bc3f3dfd62cf2a36d47c7caad56ba3103bea7e2/topobank/analysis/models.py#L68
* Tag serializer: https://github.com/ContactEngineering/topobank/blob/7bc3f3dfd62cf2a36d47c7caad56ba3103bea7e2/topobank/manager/serializers.py#L97

pastewka:
Closes by #1057

PR #1057, opened 2023-12-13
ENH: Properties for digital surface twins

IoannisNezis:
When this PR is done, digital surface twins have properties attached to them.
See #904 for more details.

- [x] Editable Properties Tab
- [x] Only editable if permissions are given 
- [x] sensible representation of properties in a database
    - [x] categorical or numerical
    - [x] name with Tagulous
    - [x] units with Tagulous
- [x] Validate Model
- [x] Serializer
- [x] Unit validation
- [x] Tests

**Optional**
- [ ] form validation


IoannisNezis:
Design example: Obsidian

pastewka:
Looks good!

IoannisNezis:
Here is what i came up with to copy this:

pastewka:
Looks good!

IoannisNezis:
I changed the way the "discard" "save" button looks:



I also want to add a [tooltip](https://getbootstrap.com/docs/4.6/components/tooltips/) to those buttons.
But this requires to add a 3rd party lib "Popper" and i wanted to get some feedback before.
Any Opinions?

pastewka:
Please keep the text for those. Icons only if there is no space.Am 20.12.2023 20:16 schrieb Ioannis Nezis ***@***.***>:
I changed the way the "discard" "save" button looks:
2023-12-20_20-14.png (view on web)
2023-12-20_20-14_1.png (view on web)
I also want to add a tooltip to those buttons.
But this requires to add a 3rd party lib "Popper" and i wanted to get some feedback before.
Any Opinions?

IoannisNezis:
Alright 👍
Is the "SAVE" intentionally all caps?
And should the edit button also be text?

topobank/manager/models.py
```diff
@@ -207,6 +207,7 @@ class Surface(models.Model, SubjectMixin):
 
     name = models.CharField(max_length=80, blank=True)
     creator = models.ForeignKey(User, on_delete=models.CASCADE)
+    properties = models.TextField(default="[]")
```
IoannisNezis, line 210: Currently i store properties in a TextField

topobank/frontend/manager/SurfaceProperties.vue
```diff
@@ -1,157 +1,164 @@
...
     }
-};
+    // Remove Properties with empty names of empty values
+    properties.value = properties.value.filter((property) => !(property.name === "" || property.value === ""));
+
+    // Patch request to store json String
+    // This is just a prove of concept, we should probably save the data in a safer structure...
+    saving.value = true;
+    axios.patch(props.surface_url, {
+        properties: JSON.stringify(properties.value)
```
IoannisNezis, line 56: In the TextField i store a JSON string of the properties.
This is just a temporary soloution and should be changed!

IoannisNezis:
The test fail because the api response now contains the 'properties' field.
This is a easy fix, but i would like to wait until the properties are stored properly.

pastewka:
> The test fail because the api response now contains the 'properties' field. This is a easy fix, but i would like to wait until the properties are stored properly.

The API tests are a bit stupid anyway, it would be good if we could think of a way to simplify. It is very annoying to add those fields manually to every test.

topobank/manager/models.py
```diff
@@ -132,6 +132,56 @@ def get_users_with_perms(self):
...
+        if isinstance(value, str):
+            self.value_categorical = value
+            self.value_numerical = None
+        elif isinstance(value, float) or isinstance(value, int):
+            self.value_numerical = value
+            self.value_categorical = None
+        else:
+            raise TypeError(f"The value must be of type int, float or str, got {type(value)}")
+
+    def save(self, *args, **kwargs):
```
pastewka, line 174: Please add a check that unit has to be None if value is categorical.

topobank/manager/models.py
```diff
@@ -147,6 +197,7 @@ class Surface(models.Model, SubjectMixin):
 
     name = models.CharField(max_length=80, blank=True)
     creator = models.ForeignKey(User, on_delete=models.CASCADE)
+    properties = models.ManyToManyField(Property)
```
pastewka, line 200: It this the right relationship? Isn't it a one-to-many relationship, as a surface can have many properties but a property can only belong to one surface?

pastewka: This would mean the property has a `Surface` as a `ForeignKey`. Then you can also add `on_delete=CASCADE` to delete properties when the `Surface` disappears.

topobank/manager/serializers.py
```diff
@@ -147,6 +147,30 @@ def get_permissions(self, obj):
...
+    class Meta:
+        model = Property
+        fields = ['name', 'value', 'unit']
+
+    name = serializers.CharField()
+    value = ValueField()
+    unit = serializers.CharField(allow_null=True)
+
+    def validate_value(self, value):
+        if not (isinstance(value, str) or isinstance(value, float) or isinstance(value, int)):
```
pastewka, line 169: Please also check that `unit` is None if `value` is categorical (i.e. `str`)

topobank/manager/serializers.py
```diff
@@ -147,6 +147,30 @@ def get_permissions(self, obj):
                                 for key, value in users.items() if key != current_user]}
 
 
+class ValueField(serializers.Field):
+
+    def to_representation(self, value):
```
pastewka, line 152: `django-rest-framwork` has no canned implementation of this?

pastewka (changes requested):

pastewka:
@IoannisNezis I've added a couple of comments and a (non-functional) example test in the new file `test_properties.py`.

I think you need to switch from many-to-many to a one-to-many relationship where the property has a `ForeignKey` to surface. The serializer then also needs the surface like here:

https://github.com/ContactEngineering/topobank/blob/dc540ae1856ccb6558db80a6dfab7ac42a9e504e/topobank/manager/serializers.py#L92

In the test you can then create a property like here:

https://github.com/ContactEngineering/topobank/blob/dc540ae1856ccb6558db80a6dfab7ac42a9e504e/topobank/manager/tests/utils.py#L30

topobank/manager/models.py
IoannisNezis: Yeah you are probably right.
I think i chose this because i did not know how to create the serialize-field for `properties` in the `Surface` serializer.
I'll change this.

topobank/manager/serializers.py
pastewka: Sounds good. It may not exist.

topobank/manager/models.py
pastewka: If you create a `ForeignKey`, then you get an automatic (implicit) field in the related dataset. For example, if you add a `ForeignKey` to `Surface`, then surface gets an attribute `property_set` (named automatically, which can be changed but I've never done this) that allows to access all properties.

pastewka: More here: https://docs.djangoproject.com/en/5.0/topics/db/examples/many_to_one/ - essentially you can simply add `topography_set` to the `Surface`-serializer

topobank/manager/models.py
IoannisNezis: I refactored this into a seperate funktion that i call at `save`.
I don't think the setter for value is the right place to check the validity of unit.

I also added the invariant: if the value is `numerical` -> `unit` is not None (but can be blank)
Here ist the docstring explaing what i want to do:
``` python
    def validate(self):
        """
        Checks the invariants of this Model.
        If any invariant is broken, a ValueError is raised

        Invariants:
        - 1. `value_categorical` or `value_numerical` are `None`
        - 2. `value_categorical` or `value_numerical` are not `None`
        This results in a 'XOR' logic and exaclty one of the value fields has to hold a value
        - 3. if `value_categorical` is not `None`, unit is `None`
        - 4. if `value_numerical` is not `None`, unit is not `None`
        This enforces the definition of a categorical values -> no units.

        Note that the `unit` field can be blank (`""`).
        `unit` == `None` and `unit` == `""`, are not the same.
        The first means that the value has no unit, the second that the value has a dimensionless unit."""
```

pastewka: Looks good. There is also a `ValidationError`, but `ValueError` also works.

topobank/manager/models.py
IoannisNezis: done.
I also changed the reverse lookup name to `peoperties`.
This is done with the `related_name` keyword.

topobank/manager/serializers.py
IoannisNezis: I checked for a while and also asked a sertan AI for help.
But this solution is still the cleanest version in my Opinion.
I'm also curious if there is a smoother way, but for now this is fine i think.
Its very readable and just a bit verbose but does the job perfectly.

pastewka: No objections!

topobank/manager/serializers.py
IoannisNezis: I added validation for this in the serializer.

It bothers me a bit that there is some code duplication now, but since we represent 'dimensionless' values differently internally and externally this is fine for me...

PR #1061, opened 2024-01-17
MAINT: increasing visibility of notification badge

IoannisNezis:
This fixes the tiny notification count in the navbar:

| Before| After |
| -- | --|
|   |   |

There were two issues:
- the code uses bootstrap-v4 components (not present)
- the font size in the custom.scss was to small (in my opinion)

Bootstrap-v5 is present because of the `bootstrap-vue-next` thing.

### DIFF
--- a/topobank/frontend/scss/custom.scss
+++ b/topobank/frontend/scss/custom.scss
@@ -618,7 +618,7 @@ body.fixed-nav.sidebar-toggled #content-wrapper {
     margin-left: 0.75rem;
     top: 0.3rem;
     font-weight: 400;
-    font-size: 0.5rem;
+    font-size: 0.7rem;
 }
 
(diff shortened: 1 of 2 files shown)


IoannisNezis:
One issue might be if the notification count is high:


Then the Badge clips in the Avatar.
This is a minor issue imho and could be ignored.

IoannisNezis:
I assumed this does not collide with the  seperation of the ui and api #1058,
since it's part of a Django template.
Is this correkt @pastewka ?

pastewka:
Thanks! All templates have been moved to the ce_ui repository. We can move this also Friday.

pastewka:
I've folded this into this PR https://github.com/ContactEngineering/ce-ui/pull/47

PR #1079, opened 2024-03-19
Remove property tagulous fields

IoannisNezis:
This PR changes the Tagulous fields of `Property` (`name` & `unit`) to TextFields.
The migration contains forward and backwards functions.
The REST api seem to require no change.


pastewka:
Hi @IoannisNezis - I would actually keep the name as a tag field. There should be no little collision there.

IoannisNezis:
Hey @pastewka,
that"s true a key should never duplicate for one DST.
I will implement a prevention in the editor to create two properties with the same key.

But i have a bad feeling about the using the Tag field without using its upsides, i think it will only introduce unexpected Behavior.
One issue i already see:
When two users create a Property with the same name at the same time we will still have a race condition.
I would not be surprised if there a more unpleasent surprises.

At the moment the Tagulous field just introduces complexity without any benefits over a simpler solution.
I would prefer migrating to a simple TextField for now and move to a Tagulous field when we actually need its features.

pastewka:
Okay, makes sense!

pastewka (approved):

Issue #547, opened 2020-11-09
Attach files to a topography


mcrot:
>   * Feature request: We should be able to store an image file with all
>     of the TEM images.
>       o Ideal would be to store it in the raw format (.dm3) which
>         contains useful metadata.
>       o Second-best would be to store it as an image file (.tif)
> 
> Yes, general attachments are on the todo list. We should support adding 
> any type of file, maybe up to a certain size.

Do you mean, you want to attach additional multiple files to a topography?

pastewka:
Yes, that's the idea.

pastewka:
Note that #331 looks like a duplicate of this.

pastewka:
Discussion with @tdbjacobs: Flat structure (no subdirectories), but multiple files. Should be include in ZIP download.

tdbjacobs:
@pastewka , @mcrot : Can I request to prioritize this enhancement? We are running into problems where people are asking to see the TEM images for our data (not just the raw (x,y) coordinates of our traced points.

pastewka:
Is this more important than #173 ?

tdbjacobs:
While I would like to say yes. The answer is no.

The vast majority of groups need 173 far more than 547!

Best,
Tevis

PR #1081, opened 2024-03-20
File Attachments

IoannisNezis:
This PR addresses #547.
When this is merged files can be attached to digital surface twins and topograpies.
Any file can be attached.
When a Surface or a topography is deleted, the attached files will be automatically deleted aswell.

Todos:
- [x] handle upload routine
- [x] solve parent indirection in serializer
- [x] handle delete routine
- [x] expand to topographies (and tags?)
- [x] handle permissions


topobank/manager/urls.py
```diff
@@ -10,6 +10,8 @@
 router.register(r'api/topography', views.TopographyViewSet, basename='topography-api')
 router.register(r'api/surface', views.SurfaceViewSet, basename='surface-api')
 router.register(r'api/property', views.PropertyViewSet, basename='property-api')
+router.register(r'api/file', views.FileManifestViewSet, basename='file-api')
+router.register(r'api/fileparent', views.FileParentViewSet, basename='fileparent-api')
```
pastewka, line 14: What does the fileparent API do? The `fileparent` is just a dispatcher between a`Topography` and a `Surface`, correct? Do we need to expose this functionality explicitly through an API?

topobank/manager/views.py
```diff
@@ -57,6 +92,8 @@ class SurfaceViewSet(mixins.CreateModelMixin,
         Prefetch('topography_set', queryset=Topography.objects.order_by('name')),
         Prefetch('properties', queryset=Property.objects.order_by('id')))
     serializer_class = SurfaceSerializer
+    # WARNING: This might be the wrong set of permissions!
```
pastewka, line 95: We should add a test that tests this...

topobank/manager/views.py
```diff
@@ -147,6 +184,43 @@ def retrieve(self, request, *args, **kwargs):
...
+        return Response(data=presigned_data)
+
+
+class FileDirectUploadFinishApi(APIView):
+
+    class InputSerializer(drf_serializers.Serializer):
+        file_id = drf_serializers.CharField()
+
+    # ToDo Permissions and Auth
+    def post(self, request):
```
pastewka, line 204: Is this the API view that is used if no S3 is present? If yes, it would be useful to also have PUT upload capability.

topobank/manager/urls.py
IoannisNezis: Yes, correct. Good catch!

(AcUtUaLlY between `Surface` or `Topgraphy` and a `FileManidest` but still very true)

I originally implemented this for debug purposes. I thought we need this relationship to "know" from the client side which parent belongs to a given attachment.
But `Surface` and `Topography` both "inject" their own URLs  into the **Attachment** component:
```
<attachments v-if="_surface != null"
    :file-parent-type="'surface'"
    :file-parent-url="_surface.url"
    :attachments="_surface.attachments"
    :permission="_permissions.current_user.permission">
</attachments>
```
This is what the `file-parent-url` prop is for.

I removed the api endpoint and everything related to it (serialier, permission, relation-fields and the viewset).

topobank/manager/views.py
IoannisNezis: No, this API view tells the backend that the upload is done.
The flow with S3 enabled is:
- start the upload (`api/upload/direct/start/`)
  - create the `FileManifest` object
- do the upload (generated upload instructions)
- finish the upload (`api/upload/direct/finish/`)
  - sets the ùpload_finished` field of the `FileManifest` object
  
  But the View bellow is the one used with S3 disabled (`FileDirectUploadLocalApi`)
  and it also uses the POST method.
 
 I have a few questions about implementing a PUT upload for the direct upload.
 I'll ask those later today in the jour fix.

PR #1106, opened 2024-07-29
FIX: preventing property duplications

### LINKED ISSUE #1105: Duplicate properties (key-value pairs)
**Describe the bug**
The API allows creation of a property that already exists. This property will the exist twice in the database.

**Expected behavior**
Property/surface pairs should be unique. Creation (POST) of a new property with the a name that already exist should update that existing property.

IoannisNezis:
If applied the Property-Serializer will throw validation errors when:

- A new property is **created** for a surface that already has a property with the same name.
- A property is **updated**(via PUT), and the new name is equal the a property of the same suface thats not the altered property

Prior to this change it was, possible to create two properties with the same name, for the same surface.

This change prevents this by invalidation in the serialization.

Fixes #1105


IoannisNezis:
I only expect a update via PUT.
Should i also handle the PATCH case?

IoannisNezis:
Unfortunately the diff is quiet messy because of the formatting...
Here is the added code in `topobank/manager/serializers.py` in the `validate` funktion of the `PropertySerializer`:

```python
        method = self.context.get("request").method
        # NOTE: On creation (POST) we need to check if the surface already has a property with the same name
        if method == "POST":
            if attrs.get("surface").properties.filter(name=attrs.get("name")).exists():
                raise serializers.ValidationError(
                    {"message": "Property names have to be unique"}
                )

        # NOTE: On update (PUT) we need to check if the surface already has a property with the same name,
        # that is not the surface we are trying to update
        if method == "PUT":
            match self.instance:
                case None:
                    # NOTE:This code should not be reachable.
                    # On update, the serializer should always hold a instance
                    pass
                case _:
                    if (
                        attrs.get("surface")
                        .properties.filter(name=attrs.get("name"))
                        .exclude(id=self.instance.id)
                        .exists()
                    ):
                        raise serializers.ValidationError(
                            {"message": "Property names have to be unique"}
                        )
```

pastewka:
Hi @IoannisNezis - Model validation is good, but I think we also need a unique constraint in the database. See here:

* https://docs.djangoproject.com/en/5.0/ref/models/constraints/#uniqueconstraint

or here

* https://docs.djangoproject.com/en/5.0/ref/models/options/#unique-together

The PR looks like you turned on `black` (which is fine with me).

IoannisNezis:
I think with the current modeling this would prevent different surfaces to have properties with the same name, right?

2024-07-30T10-16-50Z_code

This would work if the name is a tagulous tag, or in general, a separate model.
But currently its only a `TextField`.

pastewka:
You need to set a `unique_together` constraint, which means the constraint would be on combinations of `name` and `surface` fields. If you just put a constraint on the `name` field then this would prevent different surfaces to have properties of the same name as you say. 

Django supports this, I am actually not sure if this also leads to a constraint on the database side or if Django handles this.

PR #1129, opened 2024-08-23
add data_urls to analysis controller to fix spa redirect cors issue

quangthuyhoang89:
- fix spa cors issue on redirect for the PCAPlot, GCAPlot, and GPCPlots when requesting django analysis urls view.data.

### DIFF
--- a/topobank/analysis/models.py
+++ b/topobank/analysis/models.py
@@ -258,6 +258,20 @@ class Analysis(TaskStateModel):
                 "This `Analysis` does not have an id yet; the storage prefix is not yet known."
             )
         return "analyses/{}".format(self.id)
+    
+    @property
+    def storage_files(self):
+        """Return all file names in analysis id directory.
+
+        List of files names ['<file_prefix_name>/file'].
+        If storage is on filesystem, the prefix should correspond
+        to a real directory.
+        """
+        if self.id is None:
+            raise RuntimeError('This `Analysis` does not have an id yet; the storage file names is not yet known.')
+        dir_tuple = default_storage.listdir(f'analyses/{self.id}')
+        file_lists = dir_tuple[1]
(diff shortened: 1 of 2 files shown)


topobank/analysis/serializers.py
```diff
@@ -59,15 +60,23 @@ class Meta:
...
 
     def get_data_prefix(self, obj):
         return reverse('analysis:data', args=(obj.id, ''), request=self.context['request'])
+
+    def get_data_urls(self, obj):
+        file_dir = {}
+        for file_path in obj.storage_files:
+            file = file_path.split('/')[-1]
+            file_name, file_extension = file.rsplit('.', 1)
+            file_dir[file_name] = default_storage.url(file_path)
```
pastewka, line 81: Why do you split file name and extension here? I would just simply use the name for the dictionary entry.

topobank/analysis/models.py
```diff
@@ -259,6 +259,20 @@ def storage_prefix(self):
...
+    def storage_files(self):
+        """Return all file names in analysis id directory.
+
+        List of files names ['<file_prefix_name>/file'].
+        If storage is on filesystem, the prefix should correspond
+        to a real directory.
+        """
+        if self.id is None:
+            raise RuntimeError('This `Analysis` does not have an id yet; the storage file names is not yet known.')
+        dir_tuple = default_storage.listdir(f'analyses/{self.id}')
```
pastewka, line 272: This is potentially very slow, but I don't see how we can do this differently at present. We will soon have a manifest for these in the database though.

pastewka (changes requested):

quangthuyhoang89:
transfer changes to a different [PR](https://github.com/ContactEngineering/topobank/pull/1133)

PR #1136, opened 2024-09-02
return list of surfaces without tags using null or None

quangthuyhoang89:
convert null or None text to None to return list of surfaces without tags

### DIFF
--- a/topobank/manager/views.py
+++ b/topobank/manager/views.py
@@ -113,9 +113,9 @@ class SurfaceViewSet(
     def get_queryset(self):
         qs = Surface.objects.for_user(self.request.user)
         tag = self.request.query_params.get("tag", None)
-        if tag is not None or (tag == 'None' or tag == 'null'):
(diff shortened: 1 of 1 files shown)


topobank/manager/views.py
```diff
@@ -113,7 +113,9 @@ def _notify(self, instance, verb):
     def get_queryset(self):
         qs = Surface.objects.for_user(self.request.user)
         tag = self.request.query_params.get("tag", None)
-        if tag is not None:
+        if tag is not None or (tag == 'None' or tag == 'null'):
```
pastewka, line 116: Why don't you simply replace tag in the line before?

```python
if tag == 'None' or tag == 'null':
    tag = None
```

quangthuyhoang89: I refactor the code to remove the nested if statement, but just simply tag = None would skip the the condition if tag is not None and not apply the query so I created a separate condition.

PR #1152, opened 2024-12-09
MAINT: Referring to functions by a fully qualified name

pastewka:
Function are currently addressed by their display name. This PR switches to a fully qualified name of the style module.function, e.g.

```
topobank_contact.boundary_element_method
```

and a display name that is shown to the user.


pastewka:
I need to make sure this does not break anything in the deployed solution, i.e. we need to manage migrations gracefully.

Issue #1149, opened 2024-11-17
Copy attachments when making a deep copy of a surface


pastewka:
**Describe the bug**
Attachments are currently not copied. However, e.g. publishing a surface invokes deep copy to make a copy and attachments should become part of the publication.

pastewka:
This is fixed in current master

PR #1153, opened 2024-12-11
Copy attachments and properties on publish

IoannisNezis:
When publishing a surface, a clone of this surface is generated.
Currently attachments and properties are not cloned.

This PR fixes this.


IoannisNezis:
I should check out what permissions the new attachments folder has.

pastewka:
This addresses (and closes) #1149

PR #1175, opened 2025-03-11
ENH: add prediction reports list endpoint

quangthuyhoang89:


pastewka:
Hi @quangthuyhoang89 - this is the wrong place to put this. It should go into the plugin.

Issue #1170, opened 2025-03-03
Query all results for a specific workflow


pastewka:
**Is your feature request related to a problem? Please describe.**
The current backend does not allow querying results without specifying a subject (surface, topography or tag).

**Describe the solution you'd like**
It would be useful to query a list of all historic analyses that have run to display to the user.

PR #1176, opened 2025-03-14
Analysis list filter by kwargs

quangthuyhoang89:
- Filter Analysis Result View List with defined by empty subject list


pastewka:
Closes #1170

PR #1186, opened 2025-04-11
ENH: Templates for workflow parameters

quangthuyhoang89:
This PR implements a datastructure that stores arguments for workflows (the `kwargs` dictionary). This allows to template workflows for specific applications.


tests/analysis/test_models.py
```diff
@@ -18,6 +18,7 @@
     TagFactory,
     Topography1DFactory,
     TopographyAnalysisFactory,
+    WorkFlowTemplateFactory,
```
pastewka, line 21: Please decapitalize `Flow`, i.e. use `Workflow` (to be consistent with the rest of the code)

topobank/analysis/models.py
```diff
@@ -623,3 +623,62 @@ def do_submit():
...
+    # Permissions
+    #
+    permissions = models.ForeignKey(PermissionSet, on_delete=models.CASCADE, null=True)
+    #
+    # name of stored parameters
+    #
+    name = models.CharField(max_length=255)
+
+    #
+    # Parameters to be passed to the PDF document
```
pastewka, line 649: Please remove any mention of a PDF document. This is just the parameters passed to the workflow.

topobank/analysis/models.py
```diff
@@ -623,3 +623,62 @@ def do_submit():
...
+    # include long term and short term parameters across one or more
+    # prediction reports
+    #
+    parameters = models.JSONField(default=dict, blank=True)
```
pastewka, line 653: Please call this `kwargs` for consistency with the filed name in the `WorkflowResult` (currently names `Analysis`) object.

topobank/analysis/models.py
```diff
@@ -623,3 +623,62 @@ def do_submit():
...
+
+    analysis = models.ForeignKey(
```
pastewka, line 655: Why do you need this? The parameters are simply copied to the analysis, I don't this we need a reference to the analysis. Plus, there can be multiple analysis using parameteres from this template.

You need a foreign key to an `AnalysisFunction` which is not here yet. Maybe you are confusing `Analysis` (which is the results) and `AnalysisFunction` (to be renamed to `WorkflowImplementation` at some future point).

quangthuyhoang89: The original idea was that WorkflowTemplate would have a many-to-1 relationship with `Analysis` to provide parameters (both short term like operator name, date time, work order number and long term like grid size, analysis type, cutoff) to populate the pdf prediction report template. 

=Now we can definitely link to `AnalysisFunction` to populate long term parameters and have multiple different analysis connect to one workflow template, and after thinking about it, I think we can pass in short term parameters to the `Analysis` kwargs field to populate the prediction report.

pastewka: The workflows define a set of parameters that the user controls. The parameters that are actually used to run a workflow are stored in the analysis object. The template object should just serve to prepolulate those.

topobank/analysis/workflows.py
```diff
@@ -141,7 +141,9 @@ class Meta:
 
     class Parameters(pydantic.BaseModel):
         class Config:
-            extra = "forbid"
+            extra = "allow"  # Extra fields are ignored and not included in the model.
+            # extra = "forbid"  # Extra fields are forbidden and will raise an erro
+            # extra = "ignore"  # Extra fields are forbidden and will raise an erro
```
quangthuyhoang89, line 146: @pastewka updating analysis.kwargs with non-explicity initiates by Parameters causes errors if we keep it forbid and prevents from updating non-defined fields, but setting it to 'allow' breaks this unit test. anyway around this?

FAILED tests/analysis/test_rest_api.py::test_query_with_wrong_kwargs - assert 200 == 400

pastewka:
Hi @quangthuyhoang89 - this looks good to me. Is this ready to merge from your end?

quangthuyhoang89:
> Hi @quangthuyhoang89 - this looks good to me. Is this ready to merge from your end?

yes it is ready

quangthuyhoang89:
@pastewka ready for review and merge

topobank/analysis/models.py
```diff
@@ -623,3 +624,69 @@ def do_submit():
...
         transaction.on_commit(do_submit)
         return analysis
+
+
+class WorkflowTemplate(PermissionMixin, models.Model):
+    """
+    WorkflowTemplate is a model that stores the state for a workflow
+    """
+
+    class Parameters:
```
pastewka, line 634: The `Parameters` don't appear to be used anywhere. Is this a remnant of earlier implementations? If yes please cut.

pastewka (changes requested):
Hi @quangthuyhoang89 - looks good. Can you address the one comment and add an entry to the CHANGELOG.md file? Will merge then.

PR #1219, opened 2025-10-06
ENH: Return access level of current user in Surface and Topography responses

pastewka:


topobank/authorization/serializers.py
```diff
@@ -14,15 +14,62 @@ class UserPermissionSerializer(serializers.ModelSerializer):
...
         source="user", view_name="users:user-v1-detail", read_only=True
     )
     is_current_user = serializers.SerializerMethodField()
+    api = serializers.SerializerMethodField()
 
     def get_is_current_user(self, obj: UserPermission) -> bool:
-        return self.context['request'].user == obj.user
+        return self.context["request"].user == obj.user
+
+    def get_api(self, obj: UserPermission) -> dict:
```
RGClary, line 31: def get_api declared here and on line 49

topobank/authorization/serializers.py
```diff
@@ -14,15 +14,62 @@ class UserPermissionSerializer(serializers.ModelSerializer):
...
+                },
+                "required": [
+                    "add_user",
+                    "remove_user",
+                    "add_organization",
+                    "remove_organization",
+                ],
+            }
+        )
+        def get_api(self, obj: PermissionSet) -> dict:
```
RGClary, line 49: 2nd declaration of get_api. Also this is not necessary as the api endpoints are attached to the PermissionSerializer.

RGClary (changes requested):

RGClary (changes requested):

topobank/authorization/serializers.py
pastewka: Agreed. I have removed the auxiliary endpoint.

topobank/manager/v2/serializers.py
```diff
@@ -110,6 +113,7 @@ class Meta:
 
     # Everything else
     tags = TagRelatedManagerField(required=False)
+    access_level = serializers.SerializerMethodField()
```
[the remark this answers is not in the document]
pastewka: Agreed, will make that change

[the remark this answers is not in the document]
RGClary: Ok, will approve once thats in

RGClary (changes requested):

PR #1220, opened 2025-10-06
MAINT: Added stub serializers for users/orgs to reduce extra api calls

RGClary:
Created stub serializers to return User( id, username, url ) and Organization( id, name, url ). 

This allows the frontend to skip the step of making an api call for every user/org to display it.

@pastewka This will have merge conflicts with your branch.


RGClary:
Fixed the tests @pastewka. Also fixed authorize_user to 404 if user has no access instead of 403

PR #1226, opened 2025-10-15
ENH: Added creator field to WorkflowResult

RGClary:
- Added creator foreign key to workflow result
- Added custom field for creator serialization
- Added creator to ResultSerializer
- Modified WorkflowResult.submit() to set creator
- Modified prepare_dependency_tasks() to set creator

### DIFF
--- a/topobank/supplib/serializers.py
+++ b/topobank/supplib/serializers.py
@@ -252,29 +252,29 @@ class ModelRelatedField(serializers.RelatedField):
         "properties": {
             "id": {"type": "number"},
             "url": {"type": "string"},
-            "username": {"type": "string"},
+            "name": {"type": "string"},
         },
-        "required": ["id", "url", "username"],
+        "required": ["id", "url", "name"],
     }
 )
 class UserField(ModelRelatedField):
     """
     A reusable Django REST Framework related field that returns a dictionary
-    containing the user's identifier, a hyperlinked URL, and username.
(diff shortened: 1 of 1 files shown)


topobank/supplib/serializers.py
```diff
@@ -163,3 +163,59 @@ def to_representation(self, obj):
...
+        "type": "object",
+        "properties": {
+            "id": {"type": "number"},
+            "url": {"type": "string"},
+            "username": {"type": "string"},
+        },
+        "required": ["id", "url", "username"],
+    }
+)
+class CreatorField(serializers.RelatedField):
```
pastewka: Rename this to `UserField`. It is more generic than just a creator and we should use this whereever a user is returned.

pastewka (approved):

topobank/analysis/models.py
```diff
@@ -199,6 +202,11 @@ class WorkflowResult(PermissionMixin, TaskStateModel):
     # Timestamp of creation of this analysis instance
     creation_time = models.DateTimeField(auto_now_add=True)
 
+    # Creator of this analysis instance
```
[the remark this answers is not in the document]
RGClary: Yes, should be renamed now

topobank/supplib/serializers.py
RGClary: Sounds good

pastewka (approved):

PR #1235, opened 2025-11-02
BUG: Provide either `id` or `url`

pastewka:
For `RelatedModelFields` we need to accept a dictionary that has either an id or a url of the form:

`{ id: 1 }` OR `{ url: 'https://xxx' }` -- if both are provided, there is a check whether they refer to the same object

### DIFF
--- a/topobank/supplib/serializers.py
+++ b/topobank/supplib/serializers.py
@@ -649,13 +649,6 @@ class OrganizationField(ModelRelatedField):
             ]
         }
 
-    Input for deserialization::
-
-        {
-            "name": "New Project",
-            "owner_organization": {"id": 5}
-        }
-
     Notes
     -----
     - This field is pre-configured to use the 'organizations:organization-v1-detail' view name


topobank/supplib/serializers.py
```diff
@@ -253,7 +482,19 @@ def to_internal_value(self, data):
         Model instance
             The corresponding model instance.
         """
-        return self.get_queryset().get(id=data)
+        id = data.get("id", None)
+        url = data.get("url", None)
+        if id is None and url is not None:
```
RGClary: Why do we need a URL when getting an internal value? The ID should always be passed in as the object identifier

pastewka: There are some endpoints in the SPA that use the URL rather than the id as a resource identifier. I would like to keep the options to pass in URLs rather than ids.

topobank/supplib/serializers.py
```diff
@@ -172,27 +309,119 @@ def to_representation(self, obj):
...
+            "author": {"id": 42}
+        }
+
+    Input for deserialization (URL-based)::
+
+        {
+            "title": "New Article",
+            "author": {"url": "https://example.com/api/authors/42/"}
+        }
+
```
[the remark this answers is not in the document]
RGClary: Only thing is the example inputs and outputs don't match the inputs/outputs for this field.

```
{
    "id": 1,
    "title": "My Article",
    "author": {
        "id": 42,
        "url": "https://example.com/api/authors/42/",
        "name": "John Doe",
        "email": "john@example.com"
    }
}
```

```suggestion
Output with additional fields::
{
    "id": 42,
    "url": "https://example.com/api/authors/42/",
    "name": "John Doe",
    "email": "john@example.com"
}
```

PR #1238, opened 2025-11-05
ENH: Analysis v2 API

RGClary:
Went through and commented in areas for context and also to mark TODO areas.

The goal with these updates for v2 is to keep views slim but also return the exact data required for each endpoint.

### DIFF
--- a/topobank/taskapp/tasks.py
+++ b/topobank/taskapp/tasks.py
@@ -24,7 +24,7 @@ class ProgressRecorder:
 
     PROGRESS_STATE = "PROGRESS"
 
-    def __init__(self, task: celery.Task):
+    def __init__(self, task: celery.Task | None = None):
         self._task = task
         self._total = None
         self._message = None
@@ -56,7 +56,7 @@ class ProgressRecorder:
             total=total,
             message=message,
         ).model_dump()
-        if self._task.request.id:
+        if self._task and self._task.request.id:
             # Eager tasks don't have a request id (task id), and this leads to problems
             # updating the task state with the Django DB backend in testing.
             self._task.update_state(state=state, meta=meta)


topobank/manager/admin.py
```diff
@@ -1,16 +1,95 @@
...
+tagulous.models.fields.FakeQuerySet.__str__ = _fakequeryset_str
+
+
+class TagFieldAdminMixin:
+    """Mixin to fix TagField compatibility with Django 5.2
+    
+    If we upgrade to a tagulous version that supports Django 5.2,
+    this mixin can be removed.
+    """
+
```
RGClary: Tagulous lib we are using has some issues with Django 5.2. If support is added in the future we can remove this

topobank/authorization/permissions.py
```diff
@@ -1,18 +1,86 @@
...
+    "OPTIONS": OPTIONS,
+    "HEAD": HEAD,
+    "POST": POST,
+    "PUT": PUT,
+    "PATCH": PATCH,
+    "DELETE": DELETE,
+}
+
+
+class ObjectPermission(BasePermission):
```
RGClary: Renamed this to `ObjectPermission` so it better describes what it does.

topobank/authorization/permissions.py
```diff
@@ -1,18 +1,86 @@
...
-        )
+        try:
+            obj.authorize_user(request.user, METHOD_TO_PERM[request.method])
+            return True
+        except KeyError:
+            # Unknown HTTP method
+            return False
+
+
+class PermissionFilterBackend(BaseFilterBackend):
```
RGClary: This allows the new filters I implemented to work with our permissions. It also required adding `AuthorizedManagerQueryset` to `AuthorizedManager` to allow using `queryset.for_user(user, perm)`

topobank/authorization/models.py
```diff
@@ -123,7 +123,7 @@ def get_for_user(self, user: User):
     def grant_for_user(self, user: User, allow: ViewEditFull):
         """Grant permission to user"""
         existing_permissions = self.user_permissions.filter(user=user)
-        nb_existing_permissions = len(existing_permissions)
+        nb_existing_permissions = existing_permissions.count()
```
RGClary: .count() is more efficient

topobank/analysis/v1/views.py
```diff
@@ -576,6 +576,7 @@ def memory_usage(request):
     return Response(m, status=200)
 
 
+# TODO: Delete this function. Permissions are handled in auth ViewSet now.
 @api_view(["PATCH"])
 def set_result_permissions(request, workflow_id=None):
```
RGClary: We should refactor this to call the `authorization` method while retaining the same input/output.

topobank/analysis/v2/filters.py
```diff
@@ -0,0 +1,234 @@
...
+        # Check if user has access to this plugin
+        plugins_available = Organization.objects.get_plugins_available(user)
+
+        if value in plugins_available:
+            return queryset.filter(name__icontains=value)
+
+        return queryset.none()
+
+
+class ResultViewFilterSet(FilterSet):
```
RGClary: Combined a lot of the functionality in v1 into filters for v2.

i.e. no longer has an endpoint just for `named-result` not you just query the normal results endpoint with 
`?named=True`

topobank/analysis/v2/serializers.py
```diff
@@ -0,0 +1,340 @@
...
+                    })
+
+                data["subject"] = tag
+
+        return data
+
+
+class ResultListSerializer(
+    taskapp_serializers.TaskStateModelSerializer,
+):
```
RGClary: Separate create, list, detail serializers are required if we want the API to be clean and efficient.

topobank/analysis/v2/views.py
```diff
@@ -0,0 +1,144 @@
...
+        if analysis.name:
+            return Response({"message": "Cannot renew named analysis"},
+                            status=status.HTTP_403_FORBIDDEN)
+        if not analysis.subject_dispatch.is_ready():
+            return Response(
+                {"message": f"{analysis.subject_dispatch.get_type().__name__} subject(s) not ready."},
+                status=status.HTTP_400_BAD_REQUEST
+            )
+
+        if task_state == WorkflowResult.NOTRUN:
```
RGClary: Task state is set to `NOTRUN` when created in v2. The goal with this is to allow the WR to be created -> parameters saved -> then initiate a run.

topobank/analysis/models.py
```diff
@@ -82,33 +72,59 @@ def Q(subject):
...
         query = None
-        if len(tag_ids) > 0:
-            query = models.Q(subject_dispatch__tag_id__in=tag_ids) & Q(
+
+        # Build query for tags (with user permission filtering)
+        if tag_ids:
+            q = models.Q(subject_dispatch__tag_id__in=tag_ids) & Q(
                 permissions__user_permissions__user=user
             )
-        elif isinstance(subject, Topography):
```
RGClary: BUG: `subject` was being referenced here out of scope. I fixed that, but I need to test this to make sure it works as expected.

topobank/analysis/models.py
```diff
@@ -82,33 +72,59 @@ def Q(subject):
...
-            q = models.Q(subject_dispatch__topography_id=topography_ids)
+            query = q
+
+        # Build query for topographies
+        if topography_ids:
+            q = models.Q(subject_dispatch__topography_id__in=topography_ids)
             query = query | q if query else q
-        elif isinstance(subject, Surface):
```
RGClary: Same as above

topobank/analysis/models.py
```diff
@@ -123,6 +139,32 @@ def get(self):
                 "Database corruption: All subjects appear to be None/null."
             )
 
+    def get_type(self):
+        return self.get().__class__
+    
+    def is_ready(self) -> bool:
+        """Check whether the subject is in SUCCESS state."""
```
RGClary: function to determine if a WorkflowResult can be started by checking if subject is ready.

topobank/analysis/models.py
```diff
@@ -200,13 +242,26 @@ class WorkflowResult(PermissionMixin, TaskStateModel):
...
     )
 
+    # Last user updating this WorkflowResult instance (removed reverse relation)
+    updated_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL, null=True, blank=True, on_delete=models.SET_NULL, related_name="+"
+    )
+
+    # Organization owning this WorkflowResult
+    owned_by = models.ForeignKey(
+        Organization, null=True, on_delete=models.CASCADE
```
RGClary: TODO: implement delay between org deletion and data deletion.

topobank/analysis/models.py
```diff
@@ -236,14 +298,45 @@ def save(self, *args, **kwargs):
         **kwargs : dict
             Arbitrary keyword arguments.
         """
+
+        # If the analysis has a name, we remove the subject dispatch to prevent CASCADE deletion
+        # when the subject (Tag/Topography/Surface) is deleted
+        dispatch_to_cleanup = None
+        if self.name and self.subject_dispatch:
```
RGClary: We should delete the WorkflowSubject here right? It is currently orphaned in `set_name` used by v1.

This feels like a clunky way to do it though, need to look into using a signal here.

topobank/analysis/serializers.py
```diff
@@ -126,7 +126,8 @@ class Meta:
     configuration = serializers.HyperlinkedRelatedField(
         view_name="analysis:configuration-detail", read_only=True
     )
-    creator = UserField(read_only=True)
+    creation_time = serializers.DateTimeField(source="created_at", read_only=True)
+    creator = UserField(source="created_by", read_only=True)
```
RGClary: Since I updated the model, I needed to update these fields to stay consistent with v1

topobank/manager/v2/serializers.py
```diff
@@ -53,6 +53,17 @@ class Meta:
...
+        request = self.context.get("request")
+        include_thumbnail = request.query_params.get("link_thumbnail", "false").lower()
+        if include_thumbnail in ["true", "1", "yes"]:
+            return serializers.FileField().to_representation(obj.thumbnail)
+        return reverse(
+            "files:manifest-api-detail",
+            kwargs={"pk": obj.thumbnail.id},
+            request=request,
+        ) if obj.thumbnail else None
+
```
RGClary: Returns S3 link if query_param `link_thumbnail=true` otherwise it return a link to the manifest

topobank/analysis/v2/filters.py
pastewka: Perfect!

topobank/analysis/v2/serializers.py
```diff
@@ -0,0 +1,340 @@
...
+            created_by=user,
+            updated_by=user,
+            owned_by=Organization.objects.for_user(user).first(),  # TODO: Limit to one organization per user
+            function=workflow,
+            subject_dispatch=WorkflowSubject.objects.create(validated_data.pop("subject")),
+            kwargs=validated_data.get("kwargs", {}),
+            task_state=WorkflowResult.NOTRUN
+        )
+        instance.grant_permission(user, "edit")
+
```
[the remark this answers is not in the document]
pastewka: Okay

topobank/analysis/v2/serializers.py
```diff
@@ -0,0 +1,340 @@
...
+    def get_dependencies(self, obj: WorkflowResult):
+        ret = {
+            "count": len(obj.dependencies.items()),
+            "url": reverse("analysis:result-v2-deps", kwargs={"pk": obj.id})
+        }
+        return ret
+
+
+class DependencyListSerializer(serializers.Serializer):
+    """
```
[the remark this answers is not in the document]
RGClary: Right this is not actually paginating. It is just mimicking that to keep outputs similar across the API.

topobank/supplib/pagination.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework.pagination import PageNumberPagination
```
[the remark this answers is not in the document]
pastewka: I have used `LimitOffsetPagination`

topobank/manager/v2/views.py
```diff
@@ -36,23 +41,46 @@ class SurfaceViewSet(v1.SurfaceViewSet):
...
+            'attachments',
+            'thumbnail',
+            'deepzoom',
+            'datafile',
+            'squeezed_datafile',
+        ).order_by('-creation_time')
+
+    def get_serializer_class(self):
+        if self.action == 'list':
+            return TopographyV2ListSerializer
```
pastewka: See comment above on using different serializers.

topobank/supplib/serializers.py
```diff
@@ -465,36 +465,90 @@ def to_representation(self, obj):
         }
         if self.fields:
             for field in self.fields:
-                data[field] = getattr(obj, field)
+                value = getattr(obj, field)
```
pastewka: What does this code do? Please add comments to document.

PR #1313, opened 2026-04-09
SD-616: WIP: Allow workflows to run on a set of surface objects.

RGClary:


topobank/analysis/migrations/0055_workflow_name_unique.py
```diff
@@ -0,0 +1,16 @@
...
+        ("analysis", "0054_alter_workflowresult_subject_dispatch"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="workflow",
+            name="name",
+            field=models.TextField(help_text="Internal unique identifier", unique=True),
+        ),
+    ]
```
RGClary, line 16: Ran into an issue where sometimes the workflows would get initialized twice leading to duplicates of workflows in the database.

topobank/analysis/migrations/0056_add_surfaces_m2m_and_subject_hash.py
```diff
@@ -0,0 +1,24 @@
...
+            model_name='workflowresult',
+            name='surfaces',
+            field=models.ManyToManyField(blank=True, related_name='workflow_results', to='manager.surface'),
+        ),
+        migrations.AddField(
+            model_name='workflowresult',
+            name='surfaces_hash',
+            field=models.CharField(blank=True, db_index=True, max_length=64, null=True),
+        ),
+    ]
```
RGClary, line 24: Added `surfaces` M2M field alongside the `subject_dispatch` for now.

`surfaces_hash` allows quick validation checks without a request to the DB by sorting and hashing the surface_id's.

topobank/analysis/tasks.py
```diff
@@ -186,13 +186,17 @@ def schedule_workflow(self: celery.Task, analysis_id: int, force: bool):
...
 
-    # No dependencies or all dependencies finished successfully - execute directly
+    # No dependencies or all dependencies finished successfully - execute directly.
+    # Use .apply() (synchronous) so that when schedule_workflow is used as a chord
+    # header task, the chord does not fire its callback until the actual work is done.
+    # Using .apply_async() here would cause a race condition: the chord would see
+    # schedule_workflow as "complete" while execute_workflow is still running.
     _log.debug(f"{analysis_id}/{self.request.id}: Calling execute_workflow directly.")
-    execute_workflow.apply(args=(analysis.id,))
+    execute_workflow.apply(args=(analysis.id, is_dependency))
```
RGClary, line 195: I introduced a bug here while making these changes so I put a comment explaining why this needs to be `.apply()`

topobank/analysis/tasks.py
```diff
@@ -329,9 +336,26 @@ def evaluate_function(progress_recorder, timer, finished_analyses):
         tracemalloc.start()
         tracemalloc.reset_peak()
         timer = Timer(str(self.request.id))
+        on_progress = None
+        # If a callback is configured, create a progress callback.
+        callback_path = getattr(settings, 'WORKFLOW_PROGRESS_CALLBACK', None)
```
RGClary, line 341: Added this to allow a callback function to be set in settings to receive progress updates.

topobank/files/utils.py
```diff
@@ -3,4 +3,5 @@ def file_storage_path(instance, filename: str) -> str:
         raise RuntimeError(
             f"Cannot construct storage path from unsaved instance {instance}."
         )
-    return f"data-lake/{instance.id}/{filename}"
+    prefix = "uploads" if instance.kind == "raw" else "data-lake"
+    return f"{prefix}/{instance.id}/{filename}"
```
RGClary, line 7: This is unrelated to the workflows. This change is for allowing webhooks/sqs events on file upload to confirm uploads.

topobank/analysis/models.py
```diff
@@ -239,6 +239,14 @@ class WorkflowResult(PermissionMixin, TaskStateModel):
         WorkflowSubject, on_delete=cascade_or_set_null, null=True,
     )
 
+    # New surface set system (coexists with subject_dispatch during transition)
+    surfaces = models.ManyToManyField(
+        Surface, blank=True, related_name="workflow_results",
+    )
+    # Hash of surface IDs for quick lookup of existing results with the same surface set.
```
pastewka: I would prefer a generic "subject_hash" that could be extended to cover other subject types.

topobank/analysis/workflows.py
```diff
@@ -26,6 +28,30 @@
 VIZ_SERIES = "series"
 
 
+def compute_surfaces_hash(surface_ids):
+    """Compute a deterministic hash for a set of surface IDs."""
+    sorted_ids = sorted(set(surface_ids))
+    key = ",".join(str(sid) for sid in sorted_ids)
```
pastewka, line 34: I would prefix the hash with "surfaces:" so we can introduce other hash types later

pastewka (changes requested):
Looks good, only comment is that we may want a generic "subject_hash" rather than the "surfaces_hash".

topobank/analysis/models.py
RGClary: @pastewka I changed it to `subject_hash` and started looking into changing the `surfaces` M2M to a generic subject but I think that is something we need to plan out.

pastewka (approved):

PR #1320, opened 2026-06-03
SD-670: BUG: Clear task_start/end_time when setting pending

RGClary:
When a task is re-submitted, the old task_start_time/task_end_time are never reset. This fixes that so time tracking is accurate.


pastewka (changes requested):
Please add a test for this

RGClary:
> Please add a test for this

Added @pastewka

pastewka (approved):

PR #1322, opened 2026-06-04
SD-668: BUG: Fixed refresh_cache() being called multiple times per upload

RGClary:
### Problem

An uploaded topography could finish with `task_state = "success"` but **no thumbnail**. Logs for the affected measurement showed `refresh_cache` running ~3 times concurrently across different Celery workers (two overlapping in time), under two different trace IDs. The concurrent runs raced on thumbnail generation and one clobbered the other; other topographies in the same batch only survived by lucky timing.

### Root cause

**Re-entrant task dispatch.** `refresh_cache()` populates the model's *significant* fields (`data_source`, `size_x`, `size_y`, `unit`, `height_scale`, `is_periodic`) and then calls a bare `self.save()`. `Topography.save()` interprets those changed significant fields as a user edit and re-dispatches the task:

```python
# Topography.save()
if refresh_dependent_data:
    run_task(self)
```

The free `run_task()` had **no in-flight guard**, so the task that was *currently running* dispatched a second copy of itself. A duplicate S3 webhook delivery could add a third. The result was multiple concurrent `refresh_cache` executions racing on the non-atomic thumbnail write (delete-old → create-new → assign-FK, persisted only by the terminal save).

**Silent failure.** `make_thumbnail(none_on_error=True)` swallows exceptions and the outer `run_task` sets `task_state = SUCCESS` regardless, so a missing thumbnail never surfaced.

### Fix

**Bug 1 — in-flight dispatch guard (eliminates the concurrency):**
- `run_task()` now skips dispatch when the model is already in flight (`PENDING` / `PENDING_DEPENDENCIES` / `STARTED`), with a keyword-only `force=False` escape hatch. Because the outer task is `STARTED` when `refresh_cache`'s terminal `save()` re-enters `run_task`, the self-re-dispatch is suppressed — **exactly one task per upload**. This also collapses duplicate webhook deliveries.
- Added a canonical `TaskStateModel.IN_FLIGHT_STATES = (PENDING, PENDING_DEPENDENCIES, STARTED)` class attribute (no DB migration; `RETRY` deliberately excluded, consistent with the stale-task sweep).
- The `refresh_cache` management command (operator "recompute now" tool) passes `force=True` so it can still re-run a stuck/in-flight topography.

**Bug 2 — make silent failures observable (non-fatal):**
- `make_thumbnail` / `make_deepzoom` / `make_squeezed` now log at `ERROR` with `exc_info=True` (consolidating the previous WARNING + manual traceback). Generation stays resilient (`none_on_error=True` unchanged), consistent with the existing design (`test_deepzoom_creation_fails`).
- `refresh_cache` ends with a verification block that `ERROR`-logs any missing derived file when metadata is complete, so monitoring catches the case that produced this bug instead of it hiding behind `task_state=SUCCESS`.

### Files changed

| File | Change |
|------|--------|
| `topobank/taskapp/utils.py` | Module logger + in-flight guard + keyword-only `force` in `run_task()` |
| `topobank/taskapp/models.py` | `TaskStateModel.IN_FLIGHT_STATES` class attribute (no migration) |
| `topobank/manager/management/commands/refresh_cache.py` | Background branch calls `run_task(topo, force=True)` |
| `topobank/manager/models.py` | WARNING→ERROR in the three `make_*` wrappers; derived-file verification block in `refresh_cache` |
| `tests/taskapp/test_run_task_guard.py` | New: guard skip / force override / notrun dispatch, no-redispatch-when-started regression, factory path unchanged |
| `tests/manager/test_deepzoom.py` | New `test_missing_thumbnail_is_logged_but_not_fatal` |

### Testing

```
TOPOBANK_UPLOAD_METHOD=POST pytest tests/taskapp/test_run_task_guard.py tests/manager/test_deepzoom.py
```

- 7 targeted tests pass (5 new guard/regression tests + new thumbnail-logging test + existing `test_deepzoom_creation_fails`).
- The in-flight guard is transparent to the test factories: they call `refresh_cache()` with `task_state="no"` (NOTRUN), which is not an in-flight state, so behavior is unchanged and no recursion is introduced.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/topobank/manager/management/commands/refresh_cache.py
+++ b/topobank/manager/management/commands/refresh_cache.py
@@ -46,7 +46,11 @@ class Command(BaseCommand):
 
             try:
                 if options['background']:
-                    run_task(topo)
+                    # force=True: this is a deliberate operator recompute, so
+                    # re-dispatch even if a (possibly stale/stuck) task is
+                    # already in flight. Keeps parity with the non-background
+                    # branch, which calls refresh_cache() ignoring task state.
+                    run_task(topo, force=True)
                 else:
                     topo.refresh_cache()
                 num_success += 1
--- a/topobank/manager/management/commands/refresh_cache.py
+++ b/topobank/manager/management/commands/refresh_cache.py
@@ -2,6 +2,7 @@ import logging
 import traceback
 
 from django.core.management.base import BaseCommand
+from django.db import transaction
 
 from topobank.manager.models import Topography
 from topobank.taskapp.utils import run_task
@@ -50,7 +51,15 @@ class Command(BaseCommand):
                     # re-dispatch even if a (possibly stale/stuck) task is
                     # already in flight. Keeps parity with the non-background
                     # branch, which calls refresh_cache() ignoring task state.
-                    run_task(topo, force=True)
+                    #
+                    # The transaction + save() persists the pending state so an
+                    # operator sees 'pending' immediately (run_task only sets it
+                    # in memory via autosave=False) and lets run_task's on_commit
+                    # dispatch fire after the pending state is committed -- the
+                    # same pattern as Topography.ensure_task_started().
+                    with transaction.atomic():
+                        run_task(topo, force=True)
+                        topo.save()
                 else:
                     topo.refresh_cache()
                 num_success += 1
--- a/tests/manager/test_deepzoom.py
+++ b/tests/manager/test_deepzoom.py
@@ -1,3 +1,5 @@
+import logging
+
 import pytest
 
 from topobank.testing.factories import Topography2DFactory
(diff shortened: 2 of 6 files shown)


topobank/manager/management/commands/refresh_cache.py
```diff
@@ -46,7 +47,19 @@ def handle(self, *args, **options):
...
+                    # re-dispatch even if a (possibly stale/stuck) task is
+                    # already in flight. Keeps parity with the non-background
+                    # branch, which calls refresh_cache() ignoring task state.
+                    #
+                    # The transaction + save() persists the pending state so an
+                    # operator sees 'pending' immediately (run_task only sets it
+                    # in memory via autosave=False) and lets run_task's on_commit
+                    # dispatch fire after the pending state is committed -- the
+                    # same pattern as Topography.ensure_task_started().
+                    with transaction.atomic():
```
pastewka, line 60: Are you sure you want to encopsulate the whole run task in atomic commit? This sounds wrong. I would encapsulate the DB accesses within run task.

pastewka: Or would this break somewhere else because run_task is run at different locations in the code?

RGClary: Yes it needs to be here to avoid breaking things elsewhere. It guarantees that the pending state is saved to db before the dispatch without causing issues in other code.

I think this would be a good spot to look at in the workflow refactor.

pastewka (approved):

PR #1334, opened 2026-07-10
MAINT: Plumb force submit through task submission

RGClary:
## Summary

- Thread `force_submit` through `Workflow._submit_new_analysis` and `_submit_new_analysis_for_surfaces` instead of hardcoding `True` when submitting the Celery task, so a normal submission no longer forces recomputation of dependencies.
- In `prepare_dependency_tasks`, only reuse dependency results in the `SUCCESS` state; failed dependencies are always rescheduled since the failure may be transient or since-fixed.
- When re-running an existing dependency, re-stamp its `metadata["parent_workflow_result_id"]` to the parent that triggered the re-run, so task events are attributed to the triggering run rather than whichever parent originally created the row.

## Test plan

- New tests in `tests/analysis/test_tasks.py` covering dependency reuse/rescheduling and parent re-stamping
- New tests in `tests/analysis/test_models.py` and `tests/analysis/test_surface_set_workflows.py` covering `force_submit` propagation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/analysis/test_models.py
+++ b/tests/analysis/test_models.py
@@ -226,3 +226,23 @@ def test_submit_again(test_workflow):
     analysis = TopographyAnalysisFactory()
     new_analysis = analysis.submit_again()
     assert new_analysis.task_state == WorkflowResult.PENDING
+
+
+@pytest.mark.django_db
+def test_submit_dispatch_carries_callers_force_flag(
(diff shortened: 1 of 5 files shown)


RGClary:
Closing as this PR is not focused enough on the specific issue.

PR #1366, opened 2026-07-25
Refactor Topography model to pluggable Measurement with metadata schemas

pastewka:
## Summary

This is a major architectural refactoring that transforms the `Topography` model into a generic `Measurement` model with pluggable measurement types. The change decouples the persistence layer from measurement-type-specific concerns by moving all metadata into validated JSON fields and introducing a registry-based system for handling different kinds of measurements (topography maps, line scans, and future extensibility for non-height data).

### Core Model Refactoring
- **Renamed `Topography` to `Measurement`**: The model now serves as a generic record for any kind of measurement, with identity, permissions, files, and task state
- **Introduced `kind` field**: Identifies which measurement type handles this record (e.g. "topography-map", "uniform-line-scan", "nonuniform-line-scan")
- **Moved metadata to JSON fields**:
  - `metadata` JSONField: User-facing physical metadata validated by pydantic schemas
  - `file_info` JSONField: Read-only, file-derived cache written by inspection tasks
- **Updated relations**: `Surface.topography_set` → `Surface.measurements`

### Measurement Type System
- **New `topobank.measurements` app** with pluggable architecture:
  - `MeasurementType` base class: Defines how to read data, inspect files, and generate artifacts
  - `registry.py`: Centralized registration and lookup of measurement types
  - `schemas.py`: Pydantic models for metadata validation (TopographyMapMetadata, UniformLineScanMetadata, NonuniformLineScanMetadata)
  - `types.py`: Built-in implementations for height data (SurfaceTopographyType and subclasses)
  - `channels.py`: Channel identification by name with occurrence ordinals for disambiguation

### Metadata Schemas
- Replaced typed Django columns with pydantic schemas that validate JSON
- Each measurement kind has its own schema with only relevant fields (e.g. `size_y` only for maps, not line scans)
- Schemas include validators, defaults, and human-readable descriptions
- Support for coercion and dumping of metadata to/from JSON

### Data Migration
- **0084**: Rename model and relations
- **0085**: Add new JSON fields to Measurement
- **0086**: Backfill kind, metadata, file_info, and channel names from existing columns
- **0087**: Drop legacy typed columns after backfill
- **analysis/0066**: Rename `WorkflowResult.subject_topography` to `subject_measurement`

### Backward Compatibility
- Container import/export handles both legacy (flat metadata) and new (JSON) formats
- `MeasurementMeta` schema with alias choices supports reading old container format
- Management command `check_measurement_kinds` validates kind inference

### Testing Infrastructure
- New test factories: `TopographyMapFactory`, `UniformLineScanFactory`, `NonuniformLineScanFactory`
- Comprehensive tests for channel resolution, schema validation, registry behavior
- Tests for container round-tripping and legacy format import
- Tests for unregistered measurement kinds (graceful degradation)

## Notable Implementation Details

- **Channel identification**: Channels are now identified by name with optional occurrence ordinals, replacing fragile positional indices
- **Lazy filtering**: `TopobankLazySurfaceContainer` filters measurements by `yields_surface_topography` capability
- **Extensibility**: External packages can register measurement types via entry points or `AppConfig.ready()`
- **Graceful degradation**: Measurements with unregistered kinds remain usable as records even when their data cannot be interpreted
- **Significant values tracking**: Metadata fields can be marked as non-significant (don't trigger cache invalidation) via pydantic schema extras

## Merged with `main` 1.71.0

`main` released 1.71.0 while this branch was open, and two of its changes meet this refactoring. See [this comment](https://github.com/ContactEngineering/topobank/pull/1366#issuecomment-5221141195) for the reasoning; in brief:

- **The full-text search work was upstreamed.** `main`'s search-vector migration is byte-identical to this branch's `0081` and `build_search_document` is the same code, so that part of this PR is no longer new work.
- **`detrend_parameters` and `undefined_data_fraction` became `HeightFileInfo` fields** rather than the typed columns `main` added, computed in `refresh_derived_cache` from `main`'s new `manager.utils` helpers. `has_undefined_data` follows `main` in being derived from the fraction, so enabling filling no longer erases it.
- **`analysis/sizing.py`'s memory guard was ported to the JSON cache**, reading the resolutions with `KeyTextTransform` and keeping the `BigIntegerField` cast that prevents an integer overflow on large maps. This is the change most worth a close look.
- Migration numbers were shifted behind `main`'s (`0084..0087`, `analysis/0066`), `main`'s pyplot→`Figure` fix moved into `measurements/types.py` with the thumbnail code, and the changelog entry became 1.72.0.

## Files Modified/Added

- Core: `topobank/manager/models.py`, `topobank/manager/signals.py`
- New app: `topobank/measurements/` (types.py, schemas.py, registry.py, channels.py, apps.py)
- Migrations: `manager/0084`-`0087`, `analysis/0066`

https://claude.ai/code/session_011ENK9Ccwe1h2oe9xkiz5GA


pastewka:
Pushed 0fc317d3 to fix CI. Two of the red checks were genuine, one was infrastructure:

**Genuine — `tests/manager/test_container_pipeline.py` (2 failures).** These tests download the published container from contact.engineering, so they are skipped when there is no network access and did not run in my environment. Both needed updating:

- `test_full_inspection_via_task_runner` still cleared `data_source`, which no longer exists. It now clears `channel_name` — the field the inspection resolves instead — and reads the cached values from `file_info`/`metadata`.
- `test_container_import_and_datafile_read` expected an imported measurement to be readable straight away. It is not, and deliberately so: that container predates measurement kinds, and the kind follows from the selected data channel rather than from the archive's metadata, so it is unknown until the file has been inspected — the same state a freshly uploaded measurement is in. The test now asserts that contract explicitly and inspects before reading.

Worth flagging as a **behaviour change**, since it is not purely a test fix: importing a container no longer leaves its measurements immediately readable. The end state is unchanged (the first inspection determines the kind and populates everything, as it does for uploads), but the window between import and inspection is now a state in which `read()` raises rather than working off an implicit channel index. Nothing dispatches inspection at import time — before or after this change — so in practice that happens when the measurement is first viewed. If you would rather have import dispatch inspection eagerly, say so and I will add it; I left it alone because it changes the load profile of bulk imports and the old code did not do it either.

I also added the same assertions to the offline legacy-import test, so this contract is now covered by a test that runs without network access.

While looking at that I found a related sharp edge and fixed it: `TopobankLazySurfaceContainer` filters by kind, so a measurement that has not been inspected yet was **silently** omitted from the container. For a freshly imported dataset that means an analysis would quietly see zero measurements instead of failing. It now logs which measurements were skipped and how many remain.

**Infrastructure — `tests (postgres:17, …, minio, …)`.** `docker pull postgres:17` failed three times with `context deadline exceeded` against `registry-1.docker.io`; the job never reached pytest. Nothing to fix on our side, and the push should get a clean runner.

Locally: 379 passed, flake8 clean. The one local failure is `test_containers.py::test_import`, which is the same network-dependent download and passes in CI.

---

pastewka:
Merged the new `main` (1.71.0, `6f198284`) into the branch as `c1457421`. Nine files conflicted; several more merged cleanly but arrived from `main` still written against the old API. Here is what needed deciding, in the order it matters.

**The full-text search work was upstreamed, so part of this PR disappears.** `main`'s `0081_surface_search_vector_surface_surface_search_idx` is byte-identical to the one on this branch, and `build_search_document` plus the search signal handlers are the same code. Nothing to port — the diff simply collapses into `main`. Worth knowing when reading the PR: that section of the description no longer describes new work.

**Migration numbering collided twice.** `main` added `manager.0082`/`0083` for two new `Topography` columns and `analysis.0065_resultzipcontainer`. The four manager migrations here move to `0084..0087` behind `main`'s `0083`, and the subject rename to `analysis.0066`. Renumbering is safe because none of this has been applied anywhere yet.

**`main`'s two new columns became schema fields, not columns.** `detrend_parameters` and `undefined_data_fraction` are file-derived, so they are now fields of `HeightFileInfo`, computed in `refresh_derived_cache` from the helpers `main` added to `manager.utils`. The backfill migration carries them into `file_info` and the drop migration removes both columns with the rest. I kept `main`'s semantics for `has_undefined_data`, which is the substantive part: it is derived from the fraction rather than read off the filtered topography, so enabling filling no longer erases the very information the fill mode was chosen in response to.

**`analysis/sizing.py` needed real porting, and this is the one place I'd want a second opinion.** Its memory guard multiplies `resolution_x * resolution_y` in SQL to reject an analysis that cannot fit in memory — and this PR moved both resolutions into the `file_info` JSON document. The arithmetic now reads them out of the JSON with `KeyTextTransform`, keeping the `BigIntegerField` cast that stops a large map from overflowing an integer product. Two consequences:

- The filter is `file_info__resolution_x__isnull=False`. That works because `dump_metadata` uses `exclude_none=True`, so an undetermined resolution is an *absent key* rather than a JSON null — on PostgreSQL those are not the same thing. If the dumping ever starts writing explicit nulls, this filter silently stops matching and the guard quietly abstains for every measurement.
- `grid_points` reads the raw `file_info` dict rather than the typed `info` view, because a measurement of an unregistered kind has no schema to parse it with, and this module is documented to fail open rather than raise.

The guard is no longer indexable the way it was against real columns. That has no effect on the aggregate over a dataset, but the learned-coefficient query in `observed_bytes_per_point` now filters `WorkflowResult` on a JSON key across the whole memory window. It is cached for an hour and ran fine locally, but on production data volumes it is the one thing here I would watch.

Smaller items: `main`'s pyplot→`Figure` fix moved into `measurements/types.py`, where thumbnail rendering now lives (the macOS fork abort it fixes applies to the same code, just relocated); the container export's progress recorder counts measurements; `MeasurementFactory` gained `main`'s `task_state` handling, and since every factory here sets `skip_postgeneration_save` the caveat in `main`'s comment about the two factories disagreeing no longer applies; `to_dict` exports `undefined_data_fraction` for information as `main` does, while still recomputing it on import; and the changelog entry became 1.72.0 since `main` released 1.71.0 on 2026-08-03.

Locally: flake8 clean, `makemigrations --check` clean, 452 passed / 4 skipped. The network-dependent `test_containers.py::test_import` passed this time rather than being skipped, so nothing is failing here.

The PR is a draft now — I have left it that way and will not mark it ready.

---

pastewka:
Added #1383 as a follow up

PR #1386, opened 2026-08-10
API: Rename Topography to Measurement

pastewka:
PR 1 of the series splitting #1366 into separately reviewable pieces, following #1385. This is the **pure rename** — no behaviour change, no new architecture (plus one bug fix found in review, below).

**Targets the `26_measurement_model` feature branch**, not `main`. The whole series collects there and the feature branch merges to `main` in one go. It opens the `2.0.0` CHANGELOG section, because this is where the compatibility break lands.

## What changes

- `Topography` → `Measurement`
- `Surface.topography_set` → `Surface.measurements`
- `num_topographies()` → `num_measurements()`, `num_topographies_excluding_publications()` likewise
- `WorkflowResult.subject_topography` → `subject_measurement` (and `_id` / `__` lookups)
- `TopographyAnalysisFactory` → `MeasurementAnalysisFactory`, `FailedTopographyAnalysisFactory` likewise. The `subject_measurement=` kwarg rename already touches every call site, so renaming the class in the same pass avoids editing those lines twice.
- The search-index signal receivers bound to `sender=Measurement`
- `TopographyAdmin` → `MeasurementAdmin`

## Migrations

**`manager/0084`** renames the model and rewrites its `django_content_type` row **in place**. `RenameModel` does not touch that table, and notifications reach a measurement through a generic foreign key — leaving the row alone would orphan them. Updating it rather than letting `get_for_model` create a new one preserves its primary key, and with it every existing generic reference. Reversible.

**`analysis/0066`** renames the `WorkflowResult` foreign key. A rename rather than a drop-and-add: the column holds the subject of every measurement-level analysis ever computed.

## One real bug, found in review

`SubjectTypeFilter` in `analysis/admin.py` builds its lookup key as `f"subject_{value}__isnull"`, so `"topography"` was never a display label — it was half of a field name. The rename left it pointing at a column that no longer exists, and selecting the filter in the admin raised `FieldError`.

This is the class of thing a mechanical rename cannot catch: the string `subject_topography__isnull` never appears in the source, so it is invisible to a search for the old name. The filter had no test at all, which is why it broke silently — `tests/analysis/test_admin.py` now covers it, and its lookup test reads the values from `lookups()` rather than listing them, since listing them would re-encode the very assumption the bug rested on. Verified against the bug, not just for passing: restoring `"topography"` fails three of the five tests.

I swept the rest of the codebase for the same pattern afterwards. That line was the **only** place anything composes a Django field name or attribute name from a string fragment. The one adjacent family, `getattr(self, name) for name in self._significant_fields`, holds only physical column names, which this PR does not touch.

## What deliberately keeps saying "topography"

Worth stating explicitly, since a reviewer will notice each of them:

| Kept | Why |
| --- | --- |
| The subject-type tag inside `subject_hash` | Opaque, never shown to a user. Rewriting it would invalidate the hash of every existing measurement-level result. All three sites that compute it agree, which is the property that matters, and two existing tests pin the literal. |
| Index names (`topography_surface_idx`, …) | Database-internal. Renaming them is cosmetic DDL on a large table. |
| `manifest.topography_datafiles` and friends | Renaming these pulls the `files` app forward in the migration graph, which then makes the `analysis` app's `folder` field require an unrelated `AlterField`. Nothing about the model rename needs it. |
| `Topography1DFactory` / `Topography2DFactory` | These name the *dimensionality* of the data. A later PR replaces them with kind-specific factories, so renaming them here would mean renaming them twice. |

Also untouched: local variables and test names using `topography`, the container format's `topographies` key and `TopographyMeta` schema (that is the container-format PR), and `SurfaceTopography`'s own `Topography` class — `topobank/testing/fixtures.py` and `tests/manager/test_detrend_parameters.py` import it, and were excluded from the rename.

## Verification

- `flake8` clean
- `makemigrations --check --dry-run` clean, so model state and migrations agree
- **363 passed, 4 skipped**

On the rename commit alone, `--collect-only` reported the same 362 tests as `main` — nothing lost or quietly added in the substitution. That is necessary but, as the admin bug shows, not sufficient: a code path with no test cannot move the count either way, so the count matching is evidence about the substitution, not about coverage.

## Note on versioning

`pyproject.toml` uses `hatch-vcs`, so the version comes from the git tag — there is no version string to edit. Per the agreed plan the whole series lands in one `2.0.0`: this PR opens the section and the later PRs append to it, so downstream consumers (ce-ui in particular) update once rather than across two majors.

Next in the series: pluggable measurement kinds, with metadata still in typed columns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog for *TopoBank*
 
+# 2.0.0 (unreleased)
+
+- API: `Topography` is now `Measurement`. The model records a measurement --
+  identity, permissions, files and task state -- rather than something specific to
+  topography data; every measurement still holds height data, so this release only
+  renames it. `Surface.topography_set` is now `Surface.measurements`,
+  `num_topographies()` is `num_measurements()`, and
+  `WorkflowResult.subject_topography` is `subject_measurement`. There is no
+  compatibility alias, so out-of-tree code has to be updated.
+
 # 1.71.0 (2026-08-03)
 
 - ENH: A data series records the extent of its data next to the reference to its
--- a/tests/analysis/test_admin.py
+++ b/tests/analysis/test_admin.py
@@ -1,0 +1,88 @@
+"""
+Tests for the `WorkflowResult` admin.
+
+`SubjectTypeFilter` builds its lookup key by interpolating the selected value
+into ``subject_{value}__isnull``, so the value is not merely a label -- it has to
+match the name of a subject field on `WorkflowResult`. A mismatch is invisible
+until someone selects the filter in the admin, at which point the queryset raises
+`FieldError`, which is what these tests guard against.
+"""
+
+import pytest
+
+from topobank.analysis.admin import SubjectTypeFilter
+from topobank.analysis.models import WorkflowResult
+from topobank.testing.factories import (
+    MeasurementAnalysisFactory,
+    SurfaceAnalysisFactory,
+    SurfaceFactory,
+    TagAnalysisFactory,
+    TagFactory,
+)
+
+
+def filter_for(value):
+    """Build the filter with `value` selected, bypassing the request plumbing."""
+    subject_filter = SubjectTypeFilter.__new__(SubjectTypeFilter)
+    subject_filter.value = lambda: value
+    return subject_filter
+
+
+def test_every_offered_value_builds_a_valid_lookup():
+    """
+    Each value the filter offers must name an existing subject field.
+
+    The values come from `lookups` rather than being written out here, so that
+    this fails if the offered values and the model's fields drift apart -- which
+    is the whole failure mode. Needs no database: the `FieldError` is raised while
+    the lookup is resolved against the model, before any query runs.
+    """
+    for value, _label in filter_for(None).lookups(None, None):
+        queryset = filter_for(value).queryset(None, WorkflowResult.objects.all())
+        # An offered value that no branch handles would silently return the
+        # queryset untouched, so check the filter was actually applied.
+        assert queryset.query.where, f"{value!r} is offered but filters nothing"
+        str(queryset.query)  # forces lookup resolution
+
+
+def test_an_unknown_value_is_ignored_rather_than_raising():
+    queryset = WorkflowResult.objects.all()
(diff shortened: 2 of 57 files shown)


pastewka:
Addressed all six points from the Copilot review in f0f36d80. One of them was a real bug in this PR, and it is worth a human reviewer's attention because it was in the *suppressed* section rather than among the surfaced comments.

**`analysis/admin.py` — `SubjectTypeFilter` was broken.** The filter interpolates the selected value into `subject_{value}__isnull`, so `"topography"` was never a display label — it was half of a field name. Renaming the field left it pointing at a column that no longer exists, and selecting the filter in the admin raised `FieldError`. Fixed by renaming the value to `"measurement"`.

This is exactly the class of thing a whole-repository rename cannot catch: the lookup key never appears in the source as `subject_topography__isnull`, so it is invisible to a search for the old name, and matching test counts against `main` prove nothing about a code path with no test. The filter had no test at all, which is why it broke silently — so `tests/analysis/test_admin.py` now covers it. The lookup test reads the values from `lookups()` instead of listing them, since the failure mode is precisely the offered values drifting away from the model's fields; listing them here would have re-encoded the same assumption the bug rests on. Checked against the bug rather than only for passing: restoring `"topography"` fails three of the five tests.

The other five were cosmetic and are fixed as suggested:

- `Measurement.read`'s docstring said `SurfaceTopography.Measurement`. The rename had rewritten a reference to SurfaceTopography's *own* `Topography` class — a genuine mistake, since no such type exists. The two files that import that class were excluded from the rename, but this docstring was not.
- `TopographyAdmin` → `MeasurementAdmin`.
- `set_datafile_format` still reported "topographies" to the user; its messages and loop variable now say measurement.
- Docstrings on both renamed analysis factories.
- The double space in `FakeTopographyModel`'s docstring — that one predates this branch, fixed while the line was being touched anyway.

Test helper names that describe the *data* rather than the model are still unchanged on purpose (`FakeTopographyModel`, `Topography1DFactory`/`Topography2DFactory`), as described in the PR body.

Full suite: 363 passed, 4 skipped; flake8 and `makemigrations --check` clean.

---

RGClary (approved):

PR #1387, opened 2026-08-11
ENH: Record what kind of measurement a record holds

pastewka:
PR 2 of the series splitting #1366, on top of #1386. Targets `26_measurement_model`.

Adds `Measurement.kind` and a registry of measurement **handlers**. **Metadata deliberately stays in the typed columns** — moving it into validated JSON is PR 3. This PR moves only *behaviour*, so it can be reviewed without also reviewing a schema layer and a data migration.

## The shape

A handler owns everything that depends on what kind of data is being handled: which channels it can import, how the file is read, which derived artifacts exist. The model stops deciding that by testing field values.

```
Measurement.kind ──> registry ──> MeasurementHandler
                                    ├── claims_channel()   which kind a channel is imported as
                                    ├── read()             reader kwargs + filters
                                    ├── render_thumbnail()
                                    ├── make_deepzoom()
                                    └── capabilities: has_deepzoom, has_canonical_file,
                                                      yields_surface_topography, ...
```

`kind` is the stored discriminator, the handler is the resolved implementation, and `Meta.kind` on the handler is the same string as the column — spelled that way so the link is visible at every site:

```python
measurement.kind                 # "topography-map"  — stored, stable, in containers
measurement.handler              # TopographyMapHandler singleton
measurement.handler.Meta.kind    # "topography-map"
measurement.handler.has_deepzoom
```

Three built-in handlers cover height data (`topography-map`, `uniform-line-scan`, `nonuniform-line-scan`) and share `SurfaceTopographyHandler`, which holds everything going through `SurfaceTopography.IO`. A handler for another modality subclasses `MeasurementHandler` directly and brings its own data package.

## What moved out of the model

- **`read`** dispatches to the handler. The canonical-file shortcut, the reader kwargs and the filters live there now.
- **The thumbnail** is rendered by the handler, so the `dim == 1` / `dim == 2` branch is gone — each handler renders its own. The *storage* plumbing (which manifest to replace, which permissions to use) stays on the model, where it is identical for every kind. That split is the main judgement call in this PR and the thing most worth disagreeing with.
- **Deep Zoom and the canonical file** follow the `has_deepzoom` / `has_canonical_file` capabilities instead of `size_y is not None`, including in the completeness check at the end of `refresh_cache`.

`models.py` loses its `io`, `os.path`, `tempfile`, `matplotlib`, `PIL`, `ContentFile` and `Figure` imports as a result, which is a decent sign the move is real rather than cosmetic.

## Kind assignment, and why it is one-way

The kind is derived from the selected channel while the file is inspected. An existing kind is **never** overwritten:

```python
if self.kind is None:
    self.kind = infer_kind(channel)
```

A reader that changes its mind about a channel between releases must not silently turn stored data into a different kind of measurement. The test asserts `infer_kind` is *not called* rather than that the stored value is unchanged — a rule that merely happened to re-infer the same value would pass the weaker check while still being wrong.

A measurement with **no** recorded kind still works: `handler` derives one from the data file on demand. Importing a container builds measurements straight from the archive's metadata without ever inspecting a file, so nothing records a kind, and those measurements have to stay readable. Deriving does not write the column — inspection remains the only writer. `has_handler` deliberately does not fall back, so listing many measurements does not become one file open each.

## Migrations

**`0085`** adds the column. **`0086`** backfills it from what inspection already wrote, rather than reopening every stored data file:

- `resolution_y` is set only for two-dimensional data → topography map.
- `is_periodic_editable` is cleared only for non-uniform data (periodicity is meaningless there) → separates the two kinds of line scan.
- `resolution_x` is written on every inspection, so a null there means the file was never inspected. Those rows keep a null kind and get one when first inspected.

Three queryset updates, not row by row — this runs over every measurement in the database. Rows that already have a kind are left alone, so a plugin's measurements are not relabelled by column rules that know nothing about them.

## Graceful degradation

A measurement whose kind has no registered handler — the plugin that created it is gone — stays listable, renameable and deletable. Only reading its data raises, with a message naming the missing kind. That is tested, since it is the whole point of storing a key rather than a class reference.

## Naming

The handler was called `MeasurementType` and reached through `measurement.measurement_type` in the first two commits. Renamed after review feedback that the code was hard to read: the property stuttered, and the object was a "type" while the string identifying it was a "kind" — two words for one concept. `Meta.name` was the worst of it, since it held the kind, sat beside a `display_name` that was the actual name, and gave no hint of being the same string as the column.

Not `parser`, which was the other candidate: `SurfaceTopography`'s `ReaderBase` is what parses bytes here, a handler holds one, and most of what a handler does is not parsing.

The entry-point group changed with it, `topobank.measurement_types` → `topobank.measurement_handlers`. That string is the one name here that becomes public API once released, so this was the last free moment; nothing declares it yet.

## Verification

- `flake8` clean, `makemigrations --check --dry-run` clean
- **393 passed, 4 skipped** — 363 before, so 30 new tests; unchanged across the rename, which is the point of doing it as a separate commit

Checked against mutants rather than only for passing — each of these fails a test:

| Mutation | Fails |
| --- | --- |
| Backfill collapses the uniform/non-uniform split | `test_the_backfill_reconstructs_the_kind_from_the_columns[…nonuniform…]` |
| Backfill overwrites an existing kind | `test_the_backfill_leaves_an_existing_kind_alone` |
| Model drops the overwrite guard | `test_reinspection_does_not_reinterpret_an_existing_measurement` |
| `handler` requires a recorded kind | 3 tests, including the container round trip |

The last row is a regression CI caught after the first push: requiring a recorded kind broke container import. The local suite had missed it because the container tests need a live remote and **skip** in my environment — so the fix ships with an `export_container_zip` → `import_container_zip` round trip that needs no network.

## Known rough edge

Which fields are meaningful for which kind is still implicit — `size_y` being null is what makes a record a line scan, and `LineScanHandler.physical_sizes_of` knows to return a 1-tuple. That is exactly what PR 3 fixes by giving each kind a schema; flagging it here so it reads as deferred rather than missed.

Next: pydantic metadata schemas, the `metadata`/`file_info` JSON fields, backfill and column drop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 # 2.0.0 (unreleased)
 
+- ENH: The kind of data a measurement holds is now recorded explicitly, in
+  `Measurement.kind`, and a registry of measurement types decides how a record of
+  that kind is read and which derived artifacts it has. An external package can
+  register its own type through the `topobank.measurement_types` entry point,
+  which is what makes non-height data possible. Measurements whose kind has no
+  registered type -- because the package providing it is not installed -- stay
+  listable, downloadable and deletable; only reading their data is refused.
 - API: `Topography` is now `Measurement`. The model records a measurement --
   identity, permissions, files and task state -- rather than something specific to
   topography data; every measurement still holds height data, so this release only
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,12 @@
 # 2.0.0 (unreleased)
 
 - ENH: The kind of data a measurement holds is now recorded explicitly, in
-  `Measurement.kind`, and a registry of measurement types decides how a record of
-  that kind is read and which derived artifacts it has. An external package can
-  register its own type through the `topobank.measurement_types` entry point,
-  which is what makes non-height data possible. Measurements whose kind has no
-  registered type -- because the package providing it is not installed -- stay
-  listable, downloadable and deletable; only reading their data is refused.
+  `Measurement.kind`, and a registry of measurement handlers decides how a record
+  of that kind is read and which derived artifacts it has. An external package can
+  register its own handler through the `topobank.measurement_handlers` entry
+  point, which is what makes non-height data possible. Measurements whose kind has
+  no registered handler -- because the package providing it is not installed --
+  stay listable, downloadable and deletable; only reading their data is refused.
 - API: `Topography` is now `Measurement`. The model records a measurement --
   identity, permissions, files and task state -- rather than something specific to
   topography data; every measurement still holds height data, so this release only
--- a/test_settings.py
+++ b/test_settings.py
@@ -28,6 +28,7 @@ INSTALLED_APPS = [
     "topobank.testing.mock_auth.authorization.apps.AuthorizationAppConfig",
     "topobank.files.apps.FilesAppConfig",
     "topobank.manager.apps.ManagerAppConfig",
+    "topobank.measurements.apps.MeasurementsAppConfig",
     "topobank.analysis.apps.AnalysisAppConfig",
     "topobank.testing.mock_auth.organizations.apps.OrganizationsAppConfig",
     "topobank.properties.apps.PropertiesAppConfig",
--- a/tests/measurements/test_backfill_kind.py
+++ b/tests/measurements/test_backfill_kind.py
@@ -1,0 +1,93 @@
+"""
+Tests for the `kind` backfill in `manager.0086`.
+
+The migration reconstructs the kind of every already-inspected measurement from
+the typed columns, because reopening every stored data file would be far too slow.
+The rules it uses are tested here directly against the same column combinations,
+so that a change to either the rules or the columns they read has to be
+deliberate.
+"""
+
+import importlib
+
+import pytest
+
+from topobank.manager.models import Measurement
+from topobank.testing.factories import Topography1DFactory
+
+# The module name starts with a digit, so it cannot be imported with `from ...
+# import`.
+backfill_module = importlib.import_module(
+    "topobank.manager.migrations.0086_backfill_measurement_kind"
+)
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "resolution_x,resolution_y,is_periodic_editable,expected",
+    [
+        # Two-dimensional data: `resolution_y` is set only for maps.
+        (128, 128, True, "topography-map"),
+        # One-dimensional, periodicity still offered -> uniform.
+        (128, None, True, "uniform-line-scan"),
+        # One-dimensional, periodicity withdrawn -> non-uniform. Inspection clears
+        # the flag only for non-uniform data, which is what makes it usable here.
+        (128, None, False, "nonuniform-line-scan"),
+        # Never inspected: no resolution was ever written, so nothing can be said.
+        (None, None, True, None),
+    ],
+)
+def test_the_backfill_reconstructs_the_kind_from_the_columns(
+    resolution_x, resolution_y, is_periodic_editable, expected
+):
+    measurement = Topography1DFactory()
+    Measurement.objects.filter(pk=measurement.pk).update(
+        kind=None,
+        resolution_x=resolution_x,
+        resolution_y=resolution_y,
+        is_periodic_editable=is_periodic_editable,
+    )
+
+    backfill_module.backfill_kind(_Apps(), None)
+
+    assert Measurement.objects.get(pk=measurement.pk).kind == expected
+
+
+@pytest.mark.django_db
+def test_the_backfill_leaves_an_existing_kind_alone():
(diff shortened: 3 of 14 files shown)


pastewka:
All three findings addressed in ed896966. Two were real bugs, one of which predates this branch.

**Flat map divided by zero** (`adapters.py`). `render_thumbnail` normalizes heights by `max - min`, which is zero whenever every height is the same — a zeroed or synthetic surface — and the resulting NaNs went straight into the colormap. Now rendered as a single colour.

This one is **pre-existing**, not introduced here: the same two lines sit in `_render_thumbnail` on `26_measurement_model` at `models.py:1362`. Moving the code into the adapter is what put it in this diff. Fixing it here rather than separately, since the lines are already being touched.

**Kind inference did not finalize a pending upload** (`models.py`). `_infer_kind_from_datafile` read `self.datafile.file` without calling `Manifest.exists()`, which is what completes an upload. `read` does call it, but the kind is resolved *before* `read` is reached, so a measurement with an unconfirmed upload and no recorded kind would have failed during inference. Unlike the first, this one **is** mine — a consequence of the un-inspected-read fallback added earlier on this branch, and the same ordering hazard that fallback was introduced to fix, one layer up.

**Stale docstring** (the suppressed comment). `read` still claimed `MeasurementNotInspectedError` fires when the file has not been inspected, which the fallback made untrue. It now describes the actual condition: no recorded kind *and* no readable data file.

Four new tests, and I checked them against the unfixed code rather than only for passing — reverting either fix fails two of them:

| Reverted | Fails |
| --- | --- |
| Flat-map guard | `test_a_flat_map_renders_a_uniform_thumbnail`, `test_a_flat_map_at_zero_also_renders` |
| `exists()` call | `test_deriving_the_kind_finalizes_a_pending_upload`, `test_an_unreadable_data_file_is_reported_rather_than_opened` |

The flat-map tests assert a **single distinct colour** rather than merely "did not raise", because NaN through a colormap does not necessarily raise — it just produces something arbitrary, which is precisely the failure being guarded against. They also run under `np.errstate(invalid="raise", divide="raise")` so the division itself cannot pass silently. Both the constant-non-zero and all-zero cases are covered, since `max`, `min` and their difference all coincide in the second.

Full suite: 400 passed, 4 skipped (393 before). flake8 and `makemigrations --check` clean.

---

topobank/manager/migrations/0087_measurement_kind.py
RGClary, line 1: There is already a 0085 migration on main now. Could you merge main and regenerate this migration?

pastewka: Merged `main` in 7a16817b. It turned out to be more than a numbering clash, so flagging what I did rather than just saying "done".

**Two leaf nodes, not two files with the same number.** `0084_deleted_by` and this branch's `0084_rename_topography_measurement` both depend on `0083`, which Django refuses outright. And your `0084`/`0085` name `model_name="topography"`, so they *cannot* be ordered after the model rename — the two don't commute. So the graph had to be linearized rather than renumbered:

```
0084_deleted_by                         (main)
0085_rename_deletion_time_to_deleted_at (main)
0086_rename_topography_measurement      (was 0084)
0087_measurement_kind                   (was 0085)
0088_backfill_measurement_kind          (was 0086)
analysis/0066                           now depends on manager/0086
```

**This rewrites two migrations already merged into `26_measurement_model`** — the manager rename and `analysis/0066`, both from #1386. Safe, since nothing is released and nothing is deployed from the feature branch, but you should know it happened rather than find it later. It's also why I renumbered instead of `makemigrations --merge`: a merge node would have left your rename ambiguously ordered against the model rename, and picking the wrong order silently produces a migration that references a model that doesn't exist yet.

**Verified by applying, not just by importing.** `makemigrations --check` passing only means the graph matches the models. I applied the whole chain to an empty database in the order above, unapplied back to `0085`, and reapplied — the rename's `RunPython` for the content type is reversible and this exercises it.

Code conflicts were all one shape (you renamed the field, this branch renamed the model), so the resolutions take both: `Measurement.all_objects.filter(deleted_at__lt=...)`, `MeasurementAdmin` listing `deleted_at`, and `lazy_delete` keeping your new `deleted_by` argument while cascading through `measurements` instead of `topography_set`.

One thing worth your attention: **`tests/manager/test_lazy_delete.py` merged without a conflict and still broke the build.** It's a new file, so git had nothing to compare, but it imports `Topography` and collection failed. Ported to `Measurement` with the assertions unchanged. A clean merge is not evidence of a working merge when one side renames an API.

Also: `deletion_time` → `deleted_at` is a breaking rename with no CHANGELOG entry on `main` as far as I can see — you may want one independently of this PR. And `main` now has an open `1.72.0` section while this branch holds `2.0.0`; I kept both, since whether they fold together is a release decision rather than a merge decision.

flake8 clean, `makemigrations --check` clean, 407 passed and 4 skipped.

---

